PR #43, opened 2022-07-20
Create auto assign workflow

GrafanaWriter:
With PR #42, this PR automates assigning issues to projects. This PR uses a github action that we use in other projects and is triggered on label events and when the labels match type/docs it adds it to the technical documentation project

Additional changes

We may need support to add a new secret to this repository called GH_BOT_ACCESS_TOKEN as it is required by this action to be able to work properly


osg-grafana:
Looks like this was fixed by #144 .

PR #141, opened 2022-11-30
Reduce number of clicks to templates.

osg-grafana:
As a contributor who has read the guidelines and writes frequently, I want to access the templates quickly so I can start writing. Alternatively, I could bookmark the GitHub location instead of the guide. However, I want to keep _Writers’ Toolkit_ as my source of overall truth and would prefer to remember only that I need to go there for access to information quickly.


chri2547 (approved):
nice addition, thank you.

PR #146, opened 2022-12-01
fix video example

knylander-grafana:
Added escape characters to the example of the video so it renders properly.


alyssawada (approved):
looks good!

PR #157, opened 2022-12-16
Update guidelines for aliases considering versions

jdbaldry:

### DIFF
--- a/docs/sources/writing-guide/front-matter/index.md
+++ b/docs/sources/writing-guide/front-matter/index.md
@@ -143,7 +143,7 @@ For more detail about HTML redirects, refer to [HTML redirections](https://devel
 
 > **Note:** The redirect relies on first party JavaScript support which is common but not necessarily universal.
 
-## Guideline
+## Guidelines
 
 For unversioned projects, to allow content to be easily moved, include an `aliases` entry that refers to the initial published website directory.
 Hugo doesn't create a redirect `.html` file when the directory is already populated with content.
--- a/docs/sources/writing-guide/front-matter/index.md
+++ b/docs/sources/writing-guide/front-matter/index.md
@@ -145,11 +145,14 @@ For more detail about HTML redirects, refer to [HTML redirections](https://devel
 
 ## Guidelines
 
-For unversioned projects, to allow content to be easily moved, include an `aliases` entry that refers to the initial published website directory.
+### Unversioned projects:
+
+Include an `aliases` entry that refers to the initial published website directory. Adding an `aliases` entry makes it easier to move content around.
 Hugo doesn't create a redirect `.html` file when the directory is already populated with content.
 
 > **Note:** The published directory is dependent on which `content` subdirectory documentation is synced to in the website repository.
 > For example, documentation synced to a the `content/docs` directory requires the `/docs` prefix.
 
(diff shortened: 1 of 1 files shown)


chri2547 (approved):

jdbaldry:
Relates to https://github.com/grafana/technical-documentation/issues/464

docs/sources/writing-guide/front-matter/index.md
```diff
@@ -144,8 +144,11 @@ For more detail about HTML redirects, refer to [HTML redirections](https://devel
 
 ## Guideline
```
brendamuir, line 145:
```suggestion
## Guidelines
```

docs/sources/writing-guide/front-matter/index.md
```diff
@@ -144,8 +144,11 @@ For more detail about HTML redirects, refer to [HTML redirections](https://devel
...
 
-To allow content to be easily moved, include an `aliases` entry that refers to the initial published website directory.
+For unversioned projects, to allow content to be easily moved, include an `aliases` entry that refers to the initial published website directory.
 Hugo doesn't create a redirect `.html` file when the directory is already populated with content.
 
 > **Note:** The published directory is dependent on which `content` subdirectory documentation is synced to in the website repository.
 > For example, documentation synced to a the `content/docs` directory requires the `/docs` prefix.
+
+For versioned projects, do not include an `aliases` entry that refers to the initial published website directory as the version in the URL path can cause undesirable redirects.
+For example, a redirect from latest content to an old version.
```
brendamuir, line 154:
```suggestion
Unversioned projects:

Include an `aliases` entry that refers to the initial published website directory. Adding an `aliases` entry makes it easier to move content around.
Hugo doesn't create a redirect `.html` file when the directory is already populated with content.

> **Note:** The published directory is dependent on which `content` subdirectory documentation is synced to in the website repository.
> For example, documentation synced to a the `content/docs` directory requires the `/docs` prefix.

Versioned projects:
Do not include an `aliases` entry that refers to the initial published website directory.  The version in the URL path can cause undesirable redirects, such as a redirect from latest content to an old version.
```

brendamuir: Think it's worth adding why it makes it easier to move content around in the unversioned projects section? 
And what about what you should do when you create a new topic and delete the old one - or what happens with aliases when you move content around (i.e. was is their use?)

docs/sources/writing-guide/front-matter/index.md
jdbaldry: Fixed in baf3c0d

docs/sources/writing-guide/front-matter/index.md
jdbaldry: I've incorporated the suggestion in 8344fd0 and explained the benefit more clearly in 63a221d.

In terms of alias usage, I think that is already explained earlier in the page:
> Technical writers use [Hugo aliases](https://gohugo.io/content-management/urls/#aliases) to create redirects to the current page from other URLs.
> 
> If you specify `aliases` in the frontmatter, Hugo creates a directory that matches the alias entry that contains a single `.html` file.

brendamuir (approved):

PR #164, opened 2023-01-06
Update initial contribution documentation to favor Podman

jdbaldry:
Podman has the upside that containers can either be run as root or in rootless mode.


chri2547 (approved):
awesome, thank you!

PR #185, opened 2023-01-25
Add explanation of Hugo page bundles and pages

jdbaldry:
With this reference, technical writers should be able to choose when to use each source file with confidence.


jdbaldry:
I've added all of @grafana/docs-squad for visibility.

osg-grafana (approved):

docs/sources/writing-guide/documentation-structure/index.md
```diff
@@ -54,6 +54,26 @@ For more information about how to write tasks, refer to [Tasks]({{< relref "../t
...
+
+**Branch bundles** (`page/_index.md`) are required to produce page hierarchies.
+For the `/page/subpage/` URL to generated, there must first be a `page/_index.md` branch bundle source file.
+
+**Leaf bundles** (`page/index.md`) are required to conveniently and consistently bundle page assets.
+To refer to a style sheet `page/style.css` with the link `./style.css`, the link must be in a `page/index.md` leaf bundle source file.
+
+**Leaf bundles** (`page/index.md`) are also required to mount content from one part of the site to another using Hugo mounts.
+Hugo mounts use a virtual filesystem before site generation, and only directories can be mounted.
+To mount the page `/page/` to `/other/page`, there must first be a `page/index.md` leaf bundle source file.
```
jdbaldry, line 74:
```suggestion
To mount the page `/page/` to `/other/page/`, there must first be a `page/index.md` leaf bundle source file.
```

docs/sources/writing-guide/documentation-structure/index.md
```diff
@@ -54,6 +54,26 @@ For more information about how to write tasks, refer to [Tasks]({{< relref "../t
...
+
+**Leaf bundles** (`page/index.md`) are required to conveniently and consistently bundle page assets.
+To refer to a style sheet `page/style.css` with the link `./style.css`, the link must be in a `page/index.md` leaf bundle source file.
+
+**Leaf bundles** (`page/index.md`) are also required to mount content from one part of the site to another using Hugo mounts.
+Hugo mounts use a virtual filesystem before site generation, and only directories can be mounted.
+To mount the page `/page/` to `/other/page/`, there must first be a `page/index.md` leaf bundle source file.
+
+**Pages** (`page.md`) are any source files that are not leaf or branch bundles.
+It is convenient to use pages when none the behaviors of leaf or branch bundles are required as it can be easier to distinguish two source files in some text editors or IDEs.
```
chri2547, line 77:
```suggestion
It is convenient to use pages when none of the behaviors of leaf or branch bundles are required, as it can be easier to distinguish two source files in some text editors or IDEs.
```

chri2547 (approved):
Thank you for the write-up. I understand this topic much better now.

knylander-grafana:
Echoing @chri2547 sentiments! Thank you for the clear explanation.

GrafanaWriter (approved):
Thank you.

docs/sources/writing-guide/documentation-structure/index.md
```diff
@@ -54,6 +54,26 @@ For more information about how to write tasks, refer to [Tasks]({{< relref "../t
 
 > **Note:** If a directory contains multiple pages or subdirectories, it is a branch bundle, and it must include an `_index.md` file. A directory containing only one page is a leaf bundle, and the content filename must be `index.md`.
 
-## Useful links
+## Pages and page bundles
 
-- For more information about branch bundles and leaf bundles, refer to [Page bundles](https://gohugo.io/content-management/page-bundles/).
+Each web page generated by Hugo comes from one of three source files:
+- `page/_index.md`: a Hugo branch bundle
```
JStickler, line 60: I think we need to map "branch" "leaf" and "page" to the terminology we're using in the rest of this file (Parent, Child, Task).

docs/sources/writing-guide/documentation-structure/index.md
```diff
@@ -54,6 +54,26 @@ For more information about how to write tasks, refer to [Tasks]({{< relref "../t
...
+- `page/index.md`: a Hugo leaf bundle
+- `page.md`: a Hugo page
+
+Although each of the preceding examples results in the same URL (`/page/`), Hugo works with each source file differently.
+
+**Branch bundles** (`page/_index.md`) are required to produce page hierarchies.
```
JStickler, line 66:
```suggestion
**Branch bundles** (`page/_index.md`) are required to produce page hierarchies.  Branch bundle pages are the landing page or mini TOC for the files in a particular folder or directory.
```

docs/sources/writing-guide/documentation-structure/index.md
```diff
@@ -54,6 +54,26 @@ For more information about how to write tasks, refer to [Tasks]({{< relref "../t
...
+For the `/page/subpage/` URL to generated, there must first be a `page/_index.md` branch bundle source file.
+
+**Leaf bundles** (`page/index.md`) are required to conveniently and consistently bundle page assets.
```
JStickler, line 69: This is utterly confusing to me.  As a TW I shouldn't be thinking about the CSS, that should just be part of the template.  And I don't think we want community members/developers changing the CSS.

JStickler: I think we might be safer just going with the text from the Hugo page you linked to and saying "A leaf means it has no children."

docs/sources/writing-guide/documentation-structure/index.md
```diff
@@ -54,6 +54,26 @@ For more information about how to write tasks, refer to [Tasks]({{< relref "../t
...
+To refer to a style sheet `page/style.css` with the link `./style.css`, the link must be in a `page/index.md` leaf bundle source file.
+
+**Leaf bundles** (`page/index.md`) are also required to mount content from one part of the site to another using Hugo mounts.
+Hugo mounts use a virtual filesystem before site generation, and only directories can be mounted.
+To mount the page `/page/` to `/other/page/`, there must first be a `page/index.md` leaf bundle source file.
+
+**Pages** (`page.md`) are any source files that are not leaf or branch bundles.
```
JStickler, line 76: Do we have any file naming conventions on pages?  Or do we want to say that pages can have any file name?

JStickler (commented):
Couple of questions and suggestions.

docs/sources/writing-guide/documentation-structure/index.md
jdbaldry: Good suggestion, I did note that the rest of the file was using different terms. To me, "parent" and "child" are both relative terms. An intermediate branch bundle can be both a parent and a child, for instance `first/second/_index.md` in the following directory layout:
```
first/_index.md
first/second/_index.md
first/second/third.md
```

IIRC, the earlier file content talks about most branch bundles being intermediate pages and often being concepts. I guess that would mean that most "tasks" are leaf bundles or ordinary pages.

To be honest, I'm not sure I can marry the two sets of terminology cleanly because there are so many exceptions to the mapping.

docs/sources/writing-guide/documentation-structure/index.md
jdbaldry: Although the Hugo documentation describes what a leaf bundle is, it doesn't explain why one might use it. I can think of an alternative to `CSS` for bundling purpose, perhaps a static file for download makes more sense?

docs/sources/writing-guide/documentation-structure/index.md
jdbaldry: I believe file naming conventions are discussed in each of the topic types:
- https://grafana.com/docs/writers-toolkit/writing-guide/topic-types/concept/#write-a-concept-topic
- https://grafana.com/docs/writers-toolkit/writing-guide/topic-types/task/#write-a-task-topic
- https://grafana.com/docs/writers-toolkit/writing-guide/topic-types/reference/#write-a-reference-topic
- https://grafana.com/docs/writers-toolkit/writing-guide/topic-types/tutorial/#write-a-tutorial-topic

docs/sources/writing-guide/documentation-structure/index.md
jdbaldry: I think earlier in the file it is discussed that these pages tend to be concept pages though they often can be a mini ToC too.

jdbaldry:
Thanks for the review @JStickler, I think you have raised some important questions that I'm not sure how to integrate fully. If anyone has any ideas I'm all ears.

docs/sources/writing-guide/documentation-structure/index.md
JStickler: Then maybe "Branch bundle pages can be concept topics,  and may include the mini TOC for the files in a particular folder or directory."

MattDodsonEnglish:
I still don't understand why leaf bundles should be the default over page bundles. As far as I understand, it seems pages should be the default, and leaves should be set only when needed.


I guess my confusion boils down to three questions:
- **Do mounts not work recursively with page bundles?** That is, if I mount `dir/`, will Hugo not render `dir/docs.md` (and render only things like `dir/subdir/index.md`?
   - If it does mount recursively, then it sounds like we'd only need leaf bundles to mount isolated pages. Otherwise, just mount the whole dir. And if that's the case, shouldn't it be the exception? (unless someone's manually mounting every page, and I hope no one needs to do such tedium)
   -If it does _not_ mount recursively, I guess this does imply that every page should be mounted (which, again, seems horribly tedious).
- **When are we mounting?** I don't understand whether this is something that happens everywhere or in a few isolated cases. If isolated cases, shouldn't it be the exception? If everywhere, shouldn't page bundles be forbidden?
- **Why are we mounting?** I don't understand the consequence of not mounting. I thought all the pages shipped to the typical hugo `content` structure in `grafana/website`


Sorry I'm not trying to be querelous here; I really don't understand. And in the previous bullets I want to document all of my doubts and clearly state all the assumptions that I have about how things work.

I'm naturally suspicious of this design, because it imposes all the tedious manual work on the end user, the writer. For the technical writers, it's mostly a matter of annoying ux to navigate so many files "index.md", but of course the repetitive action of making the directories is also error prone. For writers who aren't on the main technical writing team, it's going to look strange and confusing. I'd think, _this looks like it requires some special technique_, and maybe forget about volunteering to document the information that I wanted to share. Although if this is necessary, of course it's much nicer to have it documented here, so thanks for that!

For now, in the k6-in-grafana prototype, I'm going to work with only page bundles, because I want to make it as easy as possible for other writers. But if it turns out that its mandatory, no problem! I wrote a bash script to automatically bundle pages into leaves. So the time to reformat it will be no longer than the time to run this script.

I think it require `gnutils` but if needed I could try to make it POSIX.

```bash
#!/bin/bash

shopt -s extglob
targetdir="$1"

#match all files that arent index.md or _index.md

nonindex="$targetdir"/"!(?(_)index)"
for i in $nonindex.md
do
    name="$targetdir"/"$(basename -s .md "$i")"
    mkdir "$name"
    mv "$name".md "$name"/index.md
done
```

Proof of concept:
<a href="https://asciinema.org/a/suWe5mmCtrZmpfqLrto7oYfy6" target="_blank"></a>

_I know real directories won't have both _index.md and index.md, but I wanted to make sure it skips everything it should.

jdbaldry:
> I still don't understand why leaf bundles should be the default over page bundles.

Leaf bundles are a kind of page bundles.

> As far as I understand, it seems pages should be the default, and leaves should be set only when needed.

The updated documentation says pages are the default when you don't require the features of leaf or branch bundles (https://github.com/grafana/writers-toolkit/pull/185/files#diff-b104c7f0eff70fb7c345eafeb4fa2ccc9aa4ade320f9dc9be574cdba0e68f5bfR77)

> * **Do mounts not work recursively with page bundles?** That is, if I mount `dir/`, will Hugo not render `dir/docs.md` (and render only things like `dir/subdir/index.md`?

Hugo mounts do mount every file within a directory including subdirectories.

>   * If it does mount recursively, then it sounds like we'd only need leaf bundles to mount isolated pages. Otherwise, just mount the whole dir. And if that's the case, shouldn't it be the exception? (unless someone's manually mounting every page, and I hope no one needs to do such tedium)

You are right, leaf bundles are only necessary for Hugo mounts to isolate the page.
I think it is worthwhile for me to add that to the explanation.

> Sorry I'm not trying to be querelous here; I really don't understand. And in the previous bullets I want to document all of my doubts and clearly state all the assumptions that I have about how things work.

Not at all! The purpose of this explanation is to get us all on the same page.
I'm certainly not _the_ authority on Hugo, I'm just writing my understanding given the docs I have read and my experience with the website repository.


> I'm naturally suspicious of this design, because it imposes all the tedious manual work on the end user, the writer. For the technical writers, it's mostly a matter of annoying ux to navigate so many files "index.md", but of course the repetitive action of making the directories is also error prone. For writers who aren't on the main technical writing team, it's going to look strange and confusing. I'd think, _this looks like it requires some special technique_, and maybe forget about volunteering to document the information that I wanted to share. Although if this is necessary, of course it's much nicer to have it documented here, so thanks for that!

Unfortunately all file paths are prone to typos, I don't think it is any more likely to typo a directory name than a regular file name.

jdbaldry:
`leaf-blower.sh` is a great name!
 :joy:

MattDodsonEnglish:
> Leaf bundles are a kind of page bundle.

:+1: 

>The updated documentation says pages are the default when you don't require the features of leaf or branch bundles (https://github.com/grafana/writers-toolkit/pull/185/files#diff-b104c7f0eff70fb7c345eafeb4fa2ccc9aa4ade320f9dc9be574cdba0e68f5bfR77)

Oh thanks! Sorry I missed that line. I'm wondering if we could move the information to the top of the section. Something like this:

> Use pages unless you need to mount a leaf bundle. If you're unsure about whether you need to mount a leaf bundle, you probably don't need to, so **default to branches and pages**.  

What do you think?

> Unfortunately all file paths are prone to typos, I don't think it is any more likely to typo a directory name than a regular file name.

Haha, yeah that's true. But I'd say the chance of typos increases since the leaf structure requires typing a directory name and `$(dirname)/index.md`. But I guess this is not a

jdbaldry:
> Oh thanks! Sorry I missed that line. I'm wondering if we could move the information to the top of the section. Something like this:
> Use pages unless you need to mount a leaf bundle. If you're unsure about whether you need to mount a leaf bundle, you probably don't need to, so default to branches and pages.

I'm definitely in favor of moving it up through the content if it will make it less easily missed.
I'm slightly concerned about the "If you're unsure".
I would prefer that these docs were comprehensive enough that this isn't needed so I guess there is more I need to explain?

MattDodsonEnglish:
I don't want to speak for anyone else, but I can say that I have no clue about when I would want to mount a file or about why or I would do so. It sounds like it could be used to repeat pages in multiple places, but I don't understand why that wouldn't be done with a shortcode. Really I thought module mounts were more for reusable logic and shortcodes were for reusable content.

Probably I'm just ignorant about the workflow, maybe other writers in the issue can chime in.

jdbaldry:
> I don't want to speak for anyone else, but I can say that I have no clue about when I would want to mount a file or about why or I would do so. It sounds like it could be used to repeat pages in multiple places, but I don't understand why that wouldn't be done with a shortcode. Really I thought module mounts were more for reusable logic and shortcodes were for reusable content.
> 
> Probably I'm just ignorant about the workflow, maybe other writers in the issue can chime in.

Thanks for clarifying, we are definitely lacking documentation about the mechanisms to reuse content. I thought it was in our technical-documentation repository but I don't see it there so I guess I made that up. I think we need to write that down.

MattDodsonEnglish:
> so I guess I made that up

I do that all the time. :-)  Anyway you're just one person, and it's unreasonable to expect every aspect of the process to be documented perfectly at all times.

But this is the writers' toolkit, and I wonder whether it's really necessary to know this for the average writer. And I don't know if the goal is still to have the audience of these docs be external readers too, but if so, can they even do anything about module mounts? For these reasons, I suggested

> If you don't know whether you need to mount a leaf bundle, you probably don't and **default to using pages**.

That's my impression of the friendliest UX to include insider details and not exclude outsiders. But, really I don't anything about it, so if I've complely missed the point, well, it won't be the first or last time that happened.

Mainly, I think every writer will be pleased that they can default to using pages. :-)

jdbaldry:
Good points @MattDodsonEnglish. I'll get this revised with higher prominence to using pages by default and then we can merge and iterate.

PR #186, opened 2023-01-26
Contents items are also links

samjewell:
This feels more useful to me, to have contents items as links

### DIFF
--- a/docs/static/templates/multiple-tasks-template.md
+++ b/docs/static/templates/multiple-tasks-template.md
@@ -32,10 +32,10 @@ Replace any text in _italics_ with content for your procedure.
 
 _Goal of procedure_:
 
-1. Configure service
-1. Set up the environment
-1. Optional: Install libraries
-1. Verify the installation
+1. [Configure service](#configure-service)
+1. [Set up the environment](#set-up-the-environment)
+1. [Optional: Install libraries](#optional-install-libraries)
+1. [Verify the installation](#verify-the-installation)
 
 ## Before you begin
 


docs/static/templates/multiple-tasks-template.md
```diff
@@ -32,10 +32,10 @@ Replace any text in _italics_ with content for your procedure.
 
 _Goal of procedure_:
 
-1. Configure service
-1. Set up the environment
-1. Optional: Install libraries
-1. Verify the installation
+1. [Configure service](#configure-service)
```
chri2547, line 35: @knylander-grafana , I believe there was a reason why we decided not to include in-page links. Was it a maintenance consideration? And should we reconsider? Adding links is more beneficial for the user, but we need to weigh that against maintenance and potentially breaking links when headings change.

samjewell:
> we need to weigh that against maintenance and potentially breaking links when headings change.

Can't we put something into the CI (continuous integration) process to guard against this? Some kind of GitHub check, to stop us breaking links both to anchors and to pages? The potential to break links is always there, and larger than this suggestion.

samjewell:
> we need to weigh that against maintenance and potentially breaking links when headings change.

Can't we put something into the CI (continuous integration) process to guard against this? Some kind of GitHub check, to stop us breaking links both to anchors and to pages? The potential to break links is always there, and larger than this suggestion.

docs/static/templates/multiple-tasks-template.md
osg-grafana: > @knylander-grafana , I believe there was a reason why we decided not to include in-page links. Was it a maintenance consideration? And should we reconsider? Adding links is more beneficial for the user, but we need to weigh that against maintenance and potentially breaking links when headings change.

https://github.com/grafana/writers-toolkit/blob/main/docs/sources/writing-guide/useful-link-text/index.md#quantity

osg-grafana: @AnneB-SEO2, what is your take on adding these types of links within the same document?

osg-grafana:
Pinged @AnneB-SEO2 about this again this morning via a DM in case the previous notification was missed.

AnneB-SEO2:
@osg-grafana so sorry I missed your mention last week! Circled with @chri2547 for some background. These shouldn't impact SEO and if they break I don't believe they'll show up on my 404 report, which is both good and bad. It'll keep the list from expanding but I'm not sure how you'll easily find broken anchor links

docs/static/templates/multiple-tasks-template.md
```diff
@@ -32,10 +32,10 @@ Replace any text in _italics_ with content for your procedure.
 
 _Goal of procedure_:
```
osg-grafana, line 33:
```suggestion
_Goal of the procedure_:
```

docs/static/templates/multiple-tasks-template.md
osg-grafana: @jdbaldry, could we use a `{{< section >}}` or some other method here so these are auto-generated? (Also, if so, could we say "Do not include" for a given section?)

osg-grafana: > @jdbaldry, could we use a `{{< section >}}` or some other method here so these are auto-generated? (Also, if so, could we say "Do not include" for a given section?)

Oh, but we would need to number them...

jdbaldry: The `section` shortcode is more for describing the child pages rather than anchors within the page. I don't see why we couldn't have another shortcode that works for anchors. For both shortcodes, it would be possible to include a parameter to switch between unordered and ordered lists.

I'll do it for the `section` shortcode today since it won't take long.

docs/static/templates/multiple-tasks-template.md
samjewell: This is outside of the scope of this PR, and should go into a separate PR.

osg-grafana:
@jdbaldry, do you want to take an automated direction with this PR, or would you prefer that the links provided verbatim be turned into relative links and move this to done?

jdbaldry:
> @jdbaldry, do you want to take an automated direction with this PR, or would you prefer that the links provided verbatim be turned into relative links, so we can move this PR to done?

`doc-validator` works on Markdown link destinations like these anchors and also on Hugo relref shortcodes. We don't run doc-validator on these static files though. Only on those in `docs/sources`.

My opinion is that we should merge this and prioritize `doc-validator` adoption if we are concerned about breaking anchor links.

osg-grafana (approved):

PR #190, opened 2023-02-09
Remove type/docs label workflow concept page

jdbaldry:
The concept is only really valuable for Grafana employees and can only cause confusion in public documentation.


osg-grafana:
Where does this information live internally to serve the internal audience? Would it make sense to qualify that this is internally oriented, rather than externally oriented?

jdbaldry:
> Where does this information live internally to serve the internal audience? Would it make sense to qualify that this is internally oriented, rather than externally oriented?

https://github.com/grafana/technical-documentation/tree/main/docs/about-the-type-docs-label-workflow

I'm keen to remove at least one of these as they are out of sync.

brendamuir (approved):

chri2547 (approved):

PR #193, opened 2023-02-21
Fix discrepancy between link and link text

osg-grafana:


brendamuir (approved):

docs/static/templates/reference-template.md
```diff
@@ -20,9 +20,9 @@ Add an introduction to the reference.
 
 <!-- The introduction is required. Include reference information, such as extensive tables, lists, or other information that is used as support for a task. Reference topics are also designed for API information.
 
-Often reference topics are linked from _task_ topics, because they contain information the user needs in order to perform a task. 
+For guidelines about writing a reference topic, see [Reference topic](https://grafana.com/docs/writers-toolkit/writing-guide/topic-types/reference/).
```
chri2547, line 23: Should this be a relref?

chri2547 (approved):
A question for you. Thank you for the contribution!

docs/static/templates/reference-template.md
osg-grafana: Not sure, due to static content. Will test and see what happens.

osg-grafana: No, because it is within a comment and does not get rendered.

PR #200, opened 2023-03-06
Docs: Move US English to international section

osg-grafana:


JStickler:
Just moving the current vague content still doesn't explicitly state our standard spelling. I was looking for something like "We have standardized on American English spelling."

chri2547 (approved):
Thank you! LGTM!

JStickler:
I'm still baffled by this PR.  When I said, "I can't find what I'm looking for in the Writer's Toolkit" it was not because of the content organization, it was because the content that I was looking for DOES NOT EXIST.  Moving existing content does not address the lack of an explicit statement in our Style Guide that we have standardized on US / American spelling.

PR #202, opened 2023-03-07
Add linter validation to metadata page.

JStickler:
I think it's useful to know that one of the things that the docs-validator linter checks if if topic headings match the title listed in the metadata, so I'm adding it to the Writer's Toolkit.


osg-grafana (approved):

GrafanaWriter (approved):
Thank you

PR #204, opened 2023-03-08
Adding admonition code to Style Guide.

JStickler:
I just had a developer asking about how we write admonitions.  And when I found the content in the Writer's Toolkit, I realized that we didn't include examples of the markdown syntax used to create Notes, Cautions, and Warnings.

### DIFF
--- a/docs/sources/style-guide/style-conventions/index.md
+++ b/docs/sources/style-guide/style-conventions/index.md
@@ -118,7 +118,7 @@ To focus a user's attention, Grafana Labs documentation includes notes, tips, ca
 
 The most common admonition is a note. A note provides additional information that the user should be aware of.
 
-For example:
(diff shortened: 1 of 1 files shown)


docs/sources/style-guide/style-conventions/index.md
```diff
@@ -118,7 +118,7 @@ To focus a user's attention, Grafana Labs documentation includes notes, tips, ca
 
 The most common admonition is a note. A note provides additional information that the user should be aware of.
 
-For example:
+The markdown syntax for a note is `> **Note**` and will appear in the published output as shown in the following example:
```
osg-grafana, line 121:
```suggestion
The Markdown syntax for a note is `> **Note:**`. The note appears in the published output as follows:
```

osg-grafana (approved):

PR #216, opened 2023-03-30
Docs: Differentiate between Grafana (product) and Grafana Labs (company), and conform to trademark policy

osg-grafana:
https://grafana.com/trademark-policy/

"Do not use the Grafana Labs Marks in the plural or possessive form."


chri2547 (approved):
Thank you!

PR #228, opened 2023-04-14
Pin docs.mk to a specific sha to prevent diffs

trevorwhitney:
We saw a [build failure](https://drone.grafana.net/grafana/loki/22538/5/2) after this common Makefile was included in the Loki repo.

My theory here is that since `docs.mk` is pulling the latest from `main`, it can be out of sync with what's checked in to the Loki repo. As a result, after we generate our helm reference doc and run a `git diff`, it picks up the diff between the version we have checked in and the latest in `main`. As long as we pin this to a specific sha, we should avoid that problem in the future.

We had to temporarily [revert](https://github.com/grafana/loki/pull/9146) the addition of the common Makefile as it started failing CI once there was drift between the version we had checked in and `main` in this repo.

### DIFF
--- a/docs/docs.mk
+++ b/docs/docs.mk
@@ -96,4 +96,4 @@ doc-validator/%:
 	DOCS_IMAGE=$(DOC_VALIDATOR_IMAGE) DOC_VALIDATOR_INCLUDE=$(subst doc-validator/,,$@) $(PWD)/make-docs $(PROJECTS)
 
 docs.mk: ## Fetch the latest version of this Makefile from Writers' Toolkit.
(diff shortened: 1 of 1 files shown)


trevorwhitney:
I'm stuck on how to pin to a sha, as every update to the url will change the sha. Maybe a tag is better?

jdbaldry:
I think I would like to keep the convenience of pulling the latest version as part of the `docs.mk` script so that I can tell people "Just run make docs.mk". 

I'll propose two alternatives in the Loki PR that has had problems.

trevorwhitney:
closing in favor of https://github.com/grafana/writers-toolkit/pull/232

PR #232, opened 2023-04-18
Use PHONY update target to avoid causing diffs in CI

jdbaldry:
CI typically invokes `make -B` because git doesn't store timestamps. This results in changes to the working tree when invoking `make docs` if the version of `docs.mk` committed to the repo is different to one in Writers' Toolkit. Some CI tests use the exit status of `git diff` to determine if generated files have been committed and this working tree change interferes.

This PR does not remove the original source file `scripts/make-docs` because that would break existing workflows.
Once this PR has been merged and rolled out, that script can be removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 .DS_Store
-/docs/make-docs
 /docs/variables.mk.local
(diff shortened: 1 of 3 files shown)


trevorwhitney (approved):
changing this to phony sounds like a good plan to me :+1:

PR #236, opened 2023-04-24
Feat: document shortcode for admonitions

### LINKED ISSUE #231: Add admonitions syntax
The admonition syntax should be mentioned in two different places, each with a different emphasis:
- [ ] Just update the examples in "Style conventions"
  - https://grafana.com/docs/writers-toolkit/style-guide/style-conventions/#admonitions
- [ ] Document the syntax in "Short codes"
  - https://grafana.com/docs/writers-toolkit/writing-guide/shortcodes/

### LINKED ISSUE #233: Should we remove the Tip admonition?
The Style conventions document a **Tip** tooltip with the note that it's "rarely used." Indeed, some grepping shows that the tip appears somewhere between 30-50 times across all docs. Most instances are in old versions of Graphite datasource docs, too.

It's not clear where a tip ends and a note begins. Perhaps the easiest way to remove this complication is just to remove the "Tip" admonition from the style guidelines.

MattDodsonEnglish:
- Documents the `admonition` shortcode, fixes #231
- Removes the `tips` admonition, fixes #233 
- Standardizes shortcode document so shortcode names render in `code` element in headings

### DIFF
--- a/docs/sources/writing-guide/shortcodes/index.md
+++ b/docs/sources/writing-guide/shortcodes/index.md
@@ -29,7 +29,7 @@ The content of the admonition must be within opening and closing tags.
 
 | Parameter | Description                                                     | Required |
 |-----------|-----------------------------------------------------------------|----------|
-| `type`    | The type of admonition. One of `note`, `caution`, or `warning`. | yes |
+| `type`    | The type of admonition. One of `"note"`, `"caution"`, or `"warning"`. | yes |
 
 
 ### Example
--- a/docs/sources/writing-guide/shortcodes/index.md
+++ b/docs/sources/writing-guide/shortcodes/index.md
@@ -23,7 +23,7 @@ The following sections describe shortcodes available for use in Grafana Markdown
 ## `admonition` shortcode
 
 The `admonition` shortcode renders its content in a blockquote or stylized banner.
-The style depends on the admonition type as defined in the Writers' Toolkit [Style conventions]({{% relref "../../style-guide/style-conventions" %}}).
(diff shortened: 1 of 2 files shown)


jdbaldry:
Ah, that's an interesting case I should fix in the tool. 

It doesn't expect a relref format with `%`. It only expects relrefs with `<`. We basically only use the latter everywhere. I can make a fix today.

jdbaldry:
I've opened https://github.com/grafana/technical-documentation/pull/686 to improve the error message for this kind of relref. 

Though both are valid syntax, I think it's easier that we all use `<` so that I don't have to explain it everywhere :D

jdbaldry (approved):
Shortcode reference is correct. Style conventions all make sense to me.

docs/sources/writing-guide/shortcodes/index.md
```diff
@@ -20,7 +20,30 @@ The following sections describe shortcodes available for use in Grafana Markdown
...
+## `admonition` shortcode
+
+The `admonition` shortcode renders its content in a blockquote or stylized banner.
+The style depends on the admonition type as defined in the Writers' Toolkit [Style conventions]({{< relref "../../style-guide/style-conventions" >}}).
+
+The content of the admonition must be within opening and closing tags.
+
+| Parameter | Description                                                     | Required |
+|-----------|-----------------------------------------------------------------|----------|
+| `type`    | The type of admonition. One of `note`, `caution`, or `warning`. | yes |
```
jdbaldry, line 32: Nit: maybe we should emphasize that each needs to be quoted:

```suggestion
| `type`    | The type of admonition. One of `"note"`, `"caution"`, or `"warning"`. | yes |
```

PR #245, opened 2023-05-17
Document responsive-table shortcode

jdbaldry:

### DIFF
--- a/docs/sources/writing-guide/shortcodes/index.md
+++ b/docs/sources/writing-guide/shortcodes/index.md
@@ -98,6 +98,21 @@ To add a figure, insert the `figure` shortcode with the following named paramete
 {{</* figure class="float-right"  src="/static/img/docs/grafana-cloud/k8sPods.png" caption="Pod view in Grafana Kubernetes Monitoring" */>}}
 ```
 
+## `responsive-table` shortcode
+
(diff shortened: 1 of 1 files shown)


josmperez (approved):
Overall LGTM with a nit or two. This add distinguishes itself from some sort of default behavior, but it doesn't state what the default behavior is. 

Also, on the same page but not something you changed: parameter names are inconsistently formatted and none of them use our newly defined style of using italics (to match Google). You could fix this now, or we could enter an issue for it to be done later.

jdbaldry:
I'm up for making the style consistent in this PR.

I've been reading some of the Google Developer documentation style guide and I'm still not 100% certain what Markdown I should apply to each parameter.

Should I be following the recommendation in [Format placeholders](https://developers.google.com/style/placeholders)? For example, *`PLACEHOLDER`*.

josmperez:
At this point to follow our own style guide I would say put all parameter names in italics. Looking again at Google, I find they seem to disagree a bit:

https://developers.google.com/style/code-in-text  Put “query parameter names” in code text

https://developers.google.com/style/api-reference-comments “Put parameter names in italic. For example, when you refer to the parameters of a method like doSomething(Uri data, int count), italicize the names data and count.”

That said, we can always make our style guide more specific if we want to.

jdbaldry:
Thanks for taking a second look. I think I've gone through and at least made the style more consistent. Let me know what you think.

PR #254, opened 2023-06-02
Document ordered list style

jdbaldry:

### DIFF
--- a/docs/sources/style-guide/style-conventions/index.md
+++ b/docs/sources/style-guide/style-conventions/index.md
@@ -90,14 +90,45 @@ Use the following techniques to enhance the scannability of content.
 - Use short, bulleted lists.
 - Use headings to divide content.
 
-## Unordered lists
+## Lists
+
+### Ordered lists
+
+Refer to the following guidelines when you write ordered lists.
+
+Grafana Labs guidance matches the [Google Markdown style guide for lists](https://google.github.io/styleguide/docguide/style.html#lists).
+
+> Markdown is smart enough to let the resulting HTML render your numbered lists correctly.
+> For longer lists that may change, especially long nested lists, use "lazy" numbering:
+>
+> ```markdown
+> 1.  Foo.
+> 1.  Bar.
+>     1.  Foofoo.
+>     1.  Barbar.
+> 1.  Baz.
+> ```
+>
+> However, if the list is small and you don’t anticipate changing it, prefer fully numbered lists, because it’s nicer to read in source:
+>
+> ```markdown
+> 1.  Foo.
+> 2.  Bar.
+> 3.  Baz.
+> ```
+
+The resulting HTML "numbering" type for lazy numbering depends on the website style.
+Presently, all levels of nested numbering results in numbers.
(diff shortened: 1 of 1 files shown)


josmperez:
Hi Jack - My personal preference is to *not* give doc authors the choice to use either "lazy" or "fully numbered" lists, because it's easier for me to edit when there is only one correct way. I think this is our current guidance and not sure why it's changing, Is this something we can poll the TWs on?

docs/sources/style-guide/style-conventions/index.md
```diff
@@ -90,14 +90,45 @@ Use the following techniques to enhance the scannability of content.
 - Use short, bulleted lists.
 - Use headings to divide content.
 
-## Unordered lists
+## Lists
+
```
knylander-grafana, line 94: Should we include term and definition lists here?

docs/sources/style-guide/style-conventions/index.md
```diff
@@ -90,14 +90,45 @@ Use the following techniques to enhance the scannability of content.
...
 
 Refer to the following guidelines when you write unordered lists.
 
-For more guidance, refer to [Lists](https://developers.google.com/style/lists) in the [Google developer style guide](https://developers.google.com/style/).
-
 - Begin list items with a capital letter unless there is a strong reason not to. For example, when you list case-sensitive parameters.
-- If they are complete sentences, end list items with a period. If one item in a list ends with a period, then apply periods to all items in the list.
+- If they're complete sentences, end list items with a period. If one item in a list ends with a period, then apply periods to all items in the list.
+
+For more guidance, refer to [Lists](https://developers.google.com/style/lists) in the [Google developer style guide](https://developers.google.com/style/).
```
knylander-grafana, line 131:
```suggestion
For more guidance, refer to [Lists](https://developers.google.com/style/lists) in the [Google developer style guide](https://developers.google.com/style/).

### Definition lits

Definition lists often used for providing a list of a term and indented definition. For example, you can use a definition list to document commands and their meanings: 

Feature
: Creates a new feature 

Features
: Creates more than one feature
: Sample additional features

To create a definition list, add the term on a new line then add a new line with a colon (`:`) along with the definition.

 For additional information, refer to the [Definition Lists](https://www.markdownguide.org/extended-syntax/#definition-lists) from the Markdown Guide.  
```

knylander-grafana (approved):
Thank you for adding content, @jdbaldry.  I've added some suggestions to your content to include definition lists.

docs/sources/style-guide/style-conventions/index.md
```diff
@@ -90,14 +90,45 @@ Use the following techniques to enhance the scannability of content.
 - Use short, bulleted lists.
 - Use headings to divide content.
 
-## Unordered lists
+## Lists
+
+### Ordered lists
```
josmperez, line 95: I don't want this comment to be a blocker, and I believe I said it somewhere else (Slack maybe?)... Adding guidance for nested lists is awesome, but personally I would rather not give people too different paths and let them decide if their list is big enough to use one path over another.

josmperez (approved):

docs/sources/style-guide/style-conventions/index.md
jdbaldry: I'd also support that behavior but the outcome of the style roundtable was to follow this style. That being said, it wasn't brought to a poll. I'll add it for a discussion in the next meeting.

PR #261, opened 2023-06-08
[References] Fixed incorrect slashes in link example

knylander-grafana:
This PR corrects slashes in an example on references/index.md that used \instead of /.

### DIFF
--- a/docs/sources/writing-guide/references/index.md
+++ b/docs/sources/writing-guide/references/index.md
@@ -122,7 +122,7 @@ Hugo references addressed across different products' docs, such as from `/docs/g
 
 To avoid broken links in these situations on grafana.com, use regular Markdown link syntax (`[link text](/docs/repo/version/folder/file/)`) instead of Hugo references. To ensure the links work in local builds, staging environments, and the live website, you **shouldn't** use a fully qualified URL with `https://grafana.com` for links to other content on grafana.com.
 
-For cross-repository links, use a standard markdown link, with the link structured like this: `\docs\repo\page\`.
+For cross-repository links, use a standard markdown link, with the link structured like this: `/docs/repo/page`.
 
 For example:
 


docs/sources/writing-guide/references/index.md
```diff
@@ -122,7 +122,7 @@ Hugo references addressed across different products' docs, such as from `/docs/g
 
 To avoid broken links in these situations on grafana.com, use regular Markdown link syntax (`[link text](/docs/repo/version/folder/file/)`) instead of Hugo references. To ensure the links work in local builds, staging environments, and the live website, you **shouldn't** use a fully qualified URL with `https://grafana.com` for links to other content on grafana.com.
```
josmperez, line 123: I think "in these situations" is a bit unclear because the preceding text involves two distinct scenarios--a) addressing other versions, b) cross-product links. Also, I suggest defining "fully qualified URL".

josmperez (changes requested):

docs/sources/writing-guide/references/index.md
jdbaldry: I think this is a very fair point but I don't want to block this PR that fixes incorrect syntax on unrelated changes.

If you have some time to submit some changes in a separate PR, I would happily review those changes.

jdbaldry: In the mean time, would you mind removing the request for changes so that this PR can be unblocked?

josmperez (approved):

docs/sources/writing-guide/references/index.md
knylander-grafana: Thank you for the suggestions. It would be great if you'd be willing to submit a new PR. This was just a quick syntax fix, but you ahve some good points.

PR #272, opened 2023-06-29
Update list example

knylander-grafana:
Update the list example text for lazy lists based upon style roundtable.


imatwawana (approved):
LGTM!

PR #277, opened 2023-07-10
Update `doc-validator` error message documentation

jdbaldry:


JStickler:
"Links to project assets must be made from a leaf bundle (index.md file)."  

Wait, does that mean you cannot have graphics in a branch (_index.md) file?

jdbaldry:
> "Links to project assets must be made from a leaf bundle (index.md file)."
> 
> Wait, does that mean you cannot have graphics in a branch (_index.md) file?

Graphics should be linked to following the guidelines in https://grafana.com/docs/writers-toolkit/writing-guide/image-guidelines/.

jdbaldry:
Yes, previously linking to any assets from any page was an error. Now it is acceptable to link to bundle assets from a leaf page bundle. Branch bundles should probably also be supported but I haven't needed to support them yet.

jdbaldry:
This was needed to unblock https://github.com/grafana/ebpf-autoinstrument/pull/169 where they are linking to non-media assets from leaf bundles.

JStickler:
>Yes, previously linking to any assets from any page was an error. 

So am I understanding this correctly, that we had an active validation rule that would throw an error if you had images on a page?

jdbaldry:
We had an active rule that would complain if you tried to link to any file or page without using `relref` or absolute path partial URLs. You could and can still embed images with Markdown image syntax which is slightly different to link syntax.

PR #288, opened 2023-08-01
correct figure shortcode args

robbymilo:
for https://github.com/grafana/website/issues/14433

Fixes two incorrect arguments for the figure shortcode.


jdbaldry (approved):
Thank you!

PR #292, opened 2023-08-09
Add repository Vale configuration

jdbaldry:
Should facilitate use of Vale locally after installing `vale` and running `vale sync` from the root of the repository.


.gitignore
```diff
@@ -1,3 +1,4 @@
 .DS_Store
 /docs/variables.mk.local
 /node_modules/
+/vale/Google/
```
heitortsergent, line 4: Could we keep this directory in the repository, and add the Google style guide files in this PR? I think this would make it easier for people to contribute to the repository and stick to the Vale guidelines without having to remember to run `vale sync`.

heitortsergent (approved):
I suggested a change, but this is working for me now on VSCode :DDD. Thanks for looking into this @jdbaldry 🙏.

.gitignore
jdbaldry: Yeah perhaps going back to vendoring is the way to go. I'll add that back in and then ensure that the container image we build uses the vendored style.

jdbaldry: Done in [75c6f09](https://github.com/grafana/writers-toolkit/pull/292/commits/75c6f09d42ae15b732e779f355202197389f4f99).

FYI, I squashed the commits locally so you might need to reset your local branch if you have one.

PR #301, opened 2023-08-25
Add "allows to" rule

jdbaldry:
Enforces a rule where if a contributor writes "allows to" they get an error asking if they want to change it to the grammatically correct "allows you to". This is a common wording error.

**TO DO**
Before we merge this, we need to 

- [ ] Write the explanation to help the writer understand the reason for the warning. 
- [ ] That explanation must be put into the page linked in the rule, otherwise that link is useless.

We can also choose to extend the message with that information but then we have to keep them in sync.


vale/Grafana/AllowsTo.yml
```diff
@@ -0,0 +1,9 @@
+extends: substitution
+message: do you mean '%s' instead of '%s'?
```
jdbaldry, line 2:
```suggestion
message: Did you mean '%s' instead of '%s'?
```

imatwawana:
The word list isn't hasn't been implemented because Ursula and I agreed to wait until September to see if we had an automated solution before going with a manual solution. @osg-grafana, I think we can move forward with the word list. What do you think?

imatwawana:
Confirmed with @osg-grafana that we can move forward with the word list. I'll try to get that done next week so we can unblock this PR.

imatwawana:
Looking at the word list, it's not actually a logical place to put this type of an explanation. I think the [doc-validator errata page](https://grafana.com/docs/writers-toolkit/review/doc-validator/errata/) would make a lot more sense if we could generalize that page to simply "Errata for validators". What do you think @jdbaldry?

jdbaldry:
Thanks for contributing the docs @imatwawana. I think this is good to go but I'll let @osg-grafana confirm this meets her needs.

imatwawana (approved):

PR #309, opened 2023-09-01
Update `docs/shared` documentation to reflect new version inference interface

jdbaldry:
The `/docs/shared` shortcode now supports version inference in the same way as the `docs/reference` shortcode.

Additionally, the version inference behavior has been modified to look up the text within the angle bracket delimiters exactly.
Previously, the value between the brackets would be downcased and underscore separated.
This behavior is preserved for transition from legacy usage but an exact lookup will override the old lookup.

For example, with the syntax `<SOMETHING VERSION>`:

1. The default version is the version inferred from the current page.
1. If there is the front matter parameter `something_version`, the value of that parameter overrides the set version.
1. Furthermore, if there is the front matter parameter `SOMETHING VERSION`, the value of that parameter overrides the set version.


docs/sources/write/shortcodes/index.md
```diff
@@ -296,3 +289,17 @@ This Markdown renders as:
...
+
+## About version inference
+
+Version inference enables the use of absolute paths that resolve correctly, irrespective of version.
+It uses special syntax using angle bracket delimiters like `<GRAFANA VERSION>`.
+
+As a convention, use the name of the target project, with spaces but no hyphens or underscores, all upper-case.
+For example, `grafana` becomes `GRAFANA`, `grafana-cloud` becomes GRAFANA CLOUD.
+
+The special syntax `<SOMETHING VERSION>`, is inferred from the version of the page.
```
imatwawana, line 301: I don't understand this part. The person writing the shortcode has to enter all of this—what exactly is being inferred?

docs/sources/write/shortcodes/index.md
```diff
@@ -296,3 +289,17 @@ This Markdown renders as:
...
+
+You can override version inference by including additional metadata in the front matter of the file.
+To override the value of `<GRAFANA VERSION>`, set the `GRAFANA VERSION` parameter in the page's front matter.
+For example, with the front matter `GRAFANA VERSION: next`, the shortcode replaces `<GRAFANA VERSION>` with `next` irrespective of the source content version.
```
imatwawana, line 305: I think phrasing this more as the writer's intent vs what the system does under the hood might be more helpful. This wording probably isn't entirely accurate but something a long the lines of: 
```suggestion
For example, to set the version to `next` irrespective of the source content version, add the following to the front matter: `GRAFANA VERSION: next`.
```

imatwawana (commented):
Mostly looks good to me, but I had a couple questions.

docs/sources/write/shortcodes/index.md
jdbaldry: I can see why this is confusing. The value of the substitution is inferred from the version of the page.

Perhaps the following is more clear:

```suggestion
The special syntax `<SOMETHING VERSION>` is substituted by the version that is inferred from the page's URL.
```

imatwawana: Yeah that makes more sense to me

jdbaldry:
Merging this to make sure the public docs are correct.

PR #310, opened 2023-09-01
Add integrations reference to repo list

imatwawana:


docs/sources/contribute-documentation/_index.md
```diff
@@ -84,6 +84,7 @@ Some content is sourced from other projects, listed below:
 - `/docs/grafana-cloud/api-reference/http-api`: https://github.com/grafana/grafana/tree/main/docs/sources/developers/http_api
 - `/docs/grafana-cloud/incident`: https://github.com/grafana/incident/tree/main/docs/sources
 - `/docs/grafana-cloud/oncall`: https://github.com/grafana/oncall/tree/main/docs/sources
+- `/docs/grafana-cloud/monitor-infrastructure/integrations/integration-reference/`: Automatically generated
```
jdbaldry, line 87: It does have a repository but the source repository is private.

Perhaps we should include that but with a note?

```suggestion
- `/docs/grafana-cloud/monitor-infrastructure/integrations/integration-reference/`: https://github.com/grafana/cloud-onboarding/tree/main/docs/sources (Grafana Labs only)
```

imatwawana: I think adding the source repo is fine, but the whole point here is that you shouldn't actually be manually touching that that **particular part** of that repo as far as I understand, so we should explicitly say that. Unless I'm misunderstanding.

jdbaldry: Mm, good point. As far as I know their repository doesn't have any explanation for how it works.

```suggestion
- `/docs/grafana-cloud/monitor-infrastructure/integrations/integration-reference/`: Automatically generated and published from https://github.com/grafana/cloud-onboarding (Grafana Labs only).
  Integration documentation is generated from the metadata in https://github.com/grafana/cloud-onboarding/tree/main/integrations and the template file https://github.com/grafana/cloud-onboarding/blob/main/pkg/integrations-api/integrations/docs/templates/layout.tmpl.
```

imatwawana: I think this is sufficient for an external contributor or writer's perspective.

PR #311, opened 2023-09-01
Add information about relrefs

imatwawana:
Add guidance for how to set a `relref` link in a shared content

### DIFF
--- a/docs/sources/write/reuse-content/reuse-shared-content/index.md
+++ b/docs/sources/write/reuse-content/reuse-shared-content/index.md
@@ -74,6 +74,8 @@ To reuse shared content:
    ---
    ```
 
+   You can include `relref` style links in the content. When doing so, set the destination path in relation to the location of the shared content file.
+
 1. In the consuming project, use the `docs/shared` shortcode to include the shared content.
 
    To consume the file `docs/sources/shared/common-introduction.md` from the latest version of the `tempo` project documentation, the shortcode would be the following:


jdbaldry (approved):
Nice, thank you. You can also use absolute arguments to the `relref` shortcode.

imatwawana:
> Nice, thank you. You can also use absolute arguments to the `relref` shortcode.

When you say absolute arguments, do you mean a fully specified URL and/or partial URL? Or something else? If so, it seems a lot faster/easier to just use a regular markdown link than to format a `relref`.

jdbaldry:
> > Nice, thank you. You can also use absolute arguments to the `relref` shortcode.
> 
> When you say absolute arguments, do you mean a fully specified URL and/or partial URL? Or something else? If so, it seems a lot faster/easier to just use a regular markdown link than to format a `relref`.

Indeed, you can specify the file path relative to the linking file or relative to the site root. The latter is typically considered an absolute path.

It is less typing to avoid the shortcode but the reason for using it is build time link checking which makes it easier for you to know that links are broken: https://grafana.com/docs/writers-toolkit/write/references/#build-time-link-checking-with-hugo.
Without the `relref` shortcode, `make docs` can't tell you a link is broken. The only way to find that out is to follow the link and make sure it takes you to the destination, either manually, or by waiting for Anne's external crawler. I'd say it's quicker to get the feedback from the build error than to manually click each link.

PR #313, opened 2023-09-06
Fix typo (misplaced `)`)

fabrizio-grafana:
I'm not sure about this, but I think there is a misplaced `)`. This PR suggests a fix for this possible typo.

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -1,13 +1,9 @@
-StylesPath = vale
 MinAlertLevel = suggestion
-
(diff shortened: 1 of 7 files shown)


jdbaldry:
Tagging @imatwawana because you are the primary owner of these docs. 

Speaking of which, would you like to be made CODEOWNER of this documentation so that you are automatically added as a reviewer on these PRs?

imatwawana (approved):
@fabrizio-grafana Thanks for the fix!

PR #328, opened 2023-09-22
First pass at adding Windows instructions

clayton-cornell:
Add Windows-specific steps to the Vale in VS Code section of the Writer's Toolkit.


clayton-cornell:
Closing PR... something is broken with Prettier and I'm caught in a loop of adding and removing carriage returns. Will try in a new PR.

PR #332, opened 2023-09-25
Add sub-headings to shared content page

imatwawana:
When I'm using the directions on the [Reuse shared content](https://grafana.com/docs/writers-toolkit/write/reuse-content/reuse-shared-content/) page, I'm almost always adding a single file in an existing `shared` directory, not adding the directory. As a result, I'm always looking for the spot where those directions begin, halfway through the steps. This PR adds sub-headings to the page to delineate the creation of the directory from the creation of the file.


jdbaldry (approved):
Thanks!

PR #334, opened 2023-09-26
Fix instructions for placeholder variables

jdbaldry:
The example demonstrated Markdown used in a code block which wouldn't be rendered as the example intended.

The preferred Markdown syntax for italic emphasis is `_` when using `prettier` so I've updated the instructions to use that instead of `*`.

I've also fixed the name of the characters to match their unicode naming and added examples of each of them in case the name is not obvious to readers


heitortsergent (approved):
The updates LGTM, but wanted to ask about Google's guidance. It looks like they don't use the less than/greater than signs for placeholders: https://developers.google.com/style/placeholders

I'm more used to seeing placeholders with those signs than without, but I'm wondering if we should just stick with Google's style?

jdbaldry:
Probably a good discussion for the style roundtable actually.

heitortsergent:
@jdbaldry I'll add it to the list for the next style roundtable :). But I'd say feel free to merge and if we decide to change the placeholder style I can open another PR.

heitortsergent:
@jdbaldry looks like the team already discussed this before, and there's a reason around hyperlinks and dashes on why we decided to stick with the `<PLACEHODLER>` format. (thanks to @imatwawana for pointing it out)

PR #340, opened 2023-10-04
Make headings plain text

imatwawana:
On the [shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/) and [front matter](https://grafana.com/docs/writers-toolkit/write/front-matter/) pages, the code font in the headings reduces the amount of white space there is between sections and makes the pages hard to scan. Propose taking things out of code font and just capitalizing them since the examples indicate how they should be written/formatted.

### DIFF
--- a/docs/sources/write/shortcodes/index.md
+++ b/docs/sources/write/shortcodes/index.md
@@ -26,7 +26,7 @@ The Grafana shortcode templates are defined in the `layouts/shortcodes` folder o
 To request custom shortcodes, [create an issue](https://github.com/grafana/writers-toolkit/issues).
 {{% /admonition %}}
 
-## Admonition { #admonition-shortcode }
+## Admonition
 
 The `admonition` shortcode renders its content in a blockquote or stylized banner.
 The style depends on the admonition type as defined in Writers' Toolkit [Style conventions]({{< relref "../style-guide/style-conventions" >}}).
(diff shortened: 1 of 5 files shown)


docs/sources/write/shortcodes/index.md
```diff
@@ -26,7 +26,7 @@ The Grafana shortcode templates are defined in the `layouts/shortcodes` folder o
 To request custom shortcodes, [create an issue](https://github.com/grafana/writers-toolkit/issues).
 {{% /admonition %}}
 
-## `admonition` shortcode
+## Admonition { #admonition-shortcode }
```
jdbaldry, line 29: Is the heading identifier trying to preserve old bookmarks?

I like the intent but feel like it might be worth it to break it anyway so that we don't have to worry about adding this to every new heading in the future.
I think it should have always been just "Name of shortcode" rather than "`name-of-shortcode` shortcode".

jdbaldry: Oh, and as you might have found out, `doc-validator` isn't aware of this heading syntax haha.

imatwawana: Yes, the heading identifier is basically acting like an alias, but I'm okay with not preserving the old headings.

jdbaldry (approved):
Beautiful!

PR #342, opened 2023-10-05
Add backend and frontend to wordlist

jdbaldry:
@imatwawana I'm not sure if we need to add this to our public word list documentation given that it is already covered in https://developers.google.com/style/word-list#backend and https://developers.google.com/style/word-list#frontend. WDYT?

Closes https://github.com/grafana/technical-documentation/issues/261


imatwawana:
I don't think so. This is just a basic spelling thing and doesn't really require an explanation in the way that things do that are in the word list.

PR #346, opened 2023-10-10
Fix formatting of `prettier` lint ignore comment

jdbaldry:
Because of the `pre` tags (`` ` ``), the encoded string is being interpreted literally.

Thankfully we can just use the actual HTML comment text and it works as expected.

To test, run `make docs` locally and browse to http://localhost:3002/docs/writers-toolkit/write/shortcodes/#code.


josmperez (approved):
LGTM.

PR #360, opened 2023-10-25
Reduce copying and pasting with the mouse

osg-grafana:


jdbaldry:
I'm not sure I understand the PR title in relation to the patch here, would you mind clarifying the intent of this PR.

osg-grafana:
> I'm not sure I understand the PR title in relation to the patch here, would you mind clarifying the intent of this PR.

Every time that I want to use this short code, I only want the shortcode syntax, not the verbiage around it. I either have to copy and paste the syntax using a mouse, or delete the excess that does not apply to my context when I use the shortcode.

jdbaldry (approved):

PR #363, opened 2023-10-30
Docs: Add an easy-to-miss step if repo is already cloned.

osg-grafana:


heitortsergent (approved):
I'm glad we didn't spend a lot of time debugging this. 😅 :shipit:

PR #364, opened 2023-11-01
Ensure that rendering matches the example

jdbaldry:
If website styling changes, the example will be broken.

It also makes updates a case of copy-paste rather than interpretation.


jdbaldry:
@osg-grafana Are you able to take a look?

PR #367, opened 2023-11-02
Add Jaeger, OTel, OpenTelemetry, and Zipkin to dictionary

jdbaldry:


vale/dictionaries/en_US-grafana.dic
```diff
@@ -1,13 +1,16 @@
-38
+42
 API/S po:noun
 Datadog/ po:noun
 ESLint/ po:noun
 Goldmark/ po:noun
 Grafana/M po:noun
+Jaeger/ po:noun
```
JStickler, line 7: FYI - Jaeger can also be an adjective (Jaeger Operator, Jaeger client, Jaeger agent)

vale/dictionaries/en_US-grafana.dic
```diff
@@ -1,13 +1,16 @@
...
 JMeter/ po:noun
 JUnit/ po:noun
 OAuth/ po:noun
+OTel/ po:noun
```
JStickler, line 11: Same with OTel (OTel API, Otel collector, etc.)

vale/dictionaries/en_US-grafana.dic
jdbaldry: Great point! I think I have to add them as separate entries so will do that.

jdbaldry: I think there is a few other places in the word list where this needs to be added also so I've opened: https://github.com/grafana/writers-toolkit/pull/369

JStickler (approved):
LGTM

PR #370, opened 2023-11-02
Docs: Add disaggregate to Vale linter list of words

osg-grafana:
Per @GrafanaWriter, use _disaggregate_ as the antonym of _aggregate_; for example, "disaggregated metrics that have been previously aggregated"

cc @noise, @nickfuzer, @09jvilla, @colega

### DIFF
--- a/vale/Google/WordList.yml
+++ b/vale/Google/WordList.yml
@@ -46,6 +46,7 @@ swap:
   content type: media type
   curated roles: predefined roles
   data are: data is
+  deaggregate: disaggregate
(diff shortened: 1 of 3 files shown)


jdbaldry (approved):
Please review the dictionary entry which adds disaggregate as a verb and also allows disaggregates and disaggregated.

osg-grafana:
> Please review the dictionary entry which adds disaggregate as a verb and also allows disaggregates and disaggregated.

Thank you, I was wondering about that.

PR #372, opened 2023-11-02
Docs: Add relref as a noun

osg-grafana:


jdbaldry:
I've tended to wrap referring to the shortcode in backticks which Vale skips when spellchecking.

For example,

```markdown
The `relref` shortcode
```

osg-grafana:
Oh, then maybe something else in my editor is complaining about every instance of `relref`. Maybe cSpell is the culprit. I will close this issue and figure out how to configure my local application.

PR #376, opened 2023-11-08
Add rule to encourage "refer to"

jdbaldry:
Motivation is https://grafana.com/docs/writers-toolkit/write/style-guide/style-conventions/#links-and-references


josmperez:
Not sure how this is supposed to work. Shouldn't the rule be substituting "refer to" for phrases like "see" instead?

josmperez (approved):

jdbaldry:
Let's see if this is too noisy!

PR #380, opened 2023-11-13
Docs: Add semantic suggestion for "allows to"

osg-grafana:
Sometimes a person writes, "... allows to ..." to convey "... make it possible to ...".

### DIFF
--- a/vale/Grafana/AllowsTo.yml
+++ b/vale/Grafana/AllowsTo.yml
@@ -7,3 +7,4 @@ action:
   name: replace
 swap:
   allows to: allows you to
+  allows to: makes it possible to


vale/Grafana/AllowsTo.yml
```diff
@@ -7,3 +7,4 @@ action:
   name: replace
 swap:
   allows to: allows you to
+  allows to: makes it possible to
```
jdbaldry, line 10: I think the current patch results in only "makes it possible to" being proposed:

> Did you mean 'makes it possible to' instead of 'allows to'?

With the following suggestion:

```suggestion
  allows to: allows you to|makes it possible to
```

You instead get the output:

> Did you mean 'allows you to' or 'makes it possible to' instead of 'allows to'?

osg-grafana: Thank you. Was just reading documentation about how to do this. I figured it was a `|`. :)

jdbaldry (approved):

osg-grafana:
@jdbaldry, what is the rationale of having a single file for one entry?

jdbaldry:
It lets us have a unique URL which in this case points to the word list entry for allows to -- https://grafana.com/docs/writers-toolkit/word-list#allows-to

PR #381, opened 2023-11-13
Add Enterprise product acronyms

osg-grafana:
GEL stands for Grafana Enterprise Logs
GEM stands for Grafana Enterprise Metrics
GET stands for Grafana Enterprise Traces

### DIFF
--- a/vale/Grafana/Acronyms.yml
+++ b/vale/Grafana/Acronyms.yml
@@ -13,6 +13,9 @@ exceptions:
   - CLI
   - CPU
   - HTTP
+  - GEL
+  - GEM
+  - GET
   - JSON
   - SDK
   - SQL


jdbaldry (commented):
Are we at the stage where we shouldn't first introduce GEM as Grafana Enterprise Metrics (GEM)?

osg-grafana:
> Are we at the stage where we shouldn't first introduce GEM as Grafana Enterprise Metrics (GEM)?

Do you mean first instance of GEM on a page, or in every instance that it is used?

jdbaldry:
I believe this rule only fires if the acronym hasn't already been spelled out.

That is, it will not warn in:

```markdown
The product is Grafana Enterprise Metrics (GEM).

GEM is the product.
```

but it will fire for:

```markdown
The product is Grafana Enterprise Metrics.

GEM is the product.
```

osg-grafana:
Closing PR because we need to warn on any usage that does not have this spelled out at the page level.

PR #393, opened 2023-11-30
Add Vale rule for README usage

jdbaldry:


jdbaldry:
For `readme` or `Readme` it will suggest `README`.

For `readme.md` or `Readme.md` it will suggest `README.md`.

(updated description with this too).

PR #396, opened 2023-12-05
Fix aliases documentation and incorporate more feedback from 'Brunch and learn' session on the topic

jdbaldry:
- Fix bad example
- Emphasize that the current directory is the directory containing the page, not the directory of the page
- Use more easily understood terminology in table demonstration

### DIFF
--- a/docs/sources/write/front-matter/index.md
+++ b/docs/sources/write/front-matter/index.md
@@ -104,15 +104,16 @@ For an alias in the page `/docs/grafana/latest/alerting/manage-notifications/`:
 
   In this example, this is the page `/docs/grafana/latest/`.
 
-The following table demonstrates the redirect URL, **REDIRECT**, created when using the relative alias, **RELATIVE ALIAS**, in the source file for **DESTINATION**:
+The following table demonstrates how to redirect **FROM PAGE**, to **TO PAGE**.
+**RELATIVE ALIAS** is added to the front matter of **TO PAGE**.
 
-| DESTINATION                                           | RELATIVE ALIAS                           | REDIRECT                                                             |
-| ----------------------------------------------------- | ---------------------------------------- | -------------------------------------------------------------------- |
-| `/docs/grafana/latest/alerting/manage-notifications/` | `./`                                     | `/docs/grafana/latest/alerting/`                                     |
-| `/docs/grafana/latest/alerting/manage-notifications/` | `./silences/`                            | `/docs/grafana/latest/alerting/silences/`                            |
-| `/docs/grafana/latest/alerting/manage-notifications/` | `./manage-notifications/create-silence/` | `/docs/grafana/latest/alerting/manage-notifications/create-silence/` |
-| `/docs/grafana/latest/alerting/manage-notifications/` | `../`                                    | `/docs/grafana/latest/`                                              |
-| `/docs/grafana/latest/alerting/manage-notifications/` | `../old-alerting/`                       | `/docs/grafana/latest/old-alerting/`                                 |
+| FROM PAGE                                                            | TO PAGE                                               | RELATIVE ALIAS                           |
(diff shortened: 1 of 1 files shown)


docs/sources/write/front-matter/index.md
```diff
@@ -96,23 +96,24 @@ To determine the relative alias, you must first understand the meaning of the cu
 
 For an alias in the page `/docs/grafana/latest/alerting/manage-notifications/`:
 
-- The current directory element (`.`) refers to the directory containing the current page.
+- The current directory element (`.`) refers to the directory containing the current page, _not_ the directory of the current page.
```
JStickler, line 99:
```suggestion
- The current directory element (`.`) refers to the parent directory of the current page, _not_ the directory that contains the current page.  An easy way to remember is that one dot means you have moved one step back in the file path to the page.
```

docs/sources/write/front-matter/index.md
```diff
@@ -96,23 +96,24 @@ To determine the relative alias, you must first understand the meaning of the cu
...
 
-  In this example, this is the page `/docs/grafana/latest/alerting/manage-notifications/`.
+  In this example, this is the page `/docs/grafana/latest/alerting/`.
 
 - The parent directory element (`..`) refers to the parent directory of the current directory element.
```
JStickler, line 103:
```suggestion
- The parent directory element (`..`) refers to the parent directory of the current directory element. An easy way to remember is that two dots means you have moved two steps back in the file path to the page.
```

docs/sources/write/front-matter/index.md
```diff
@@ -96,23 +96,24 @@ To determine the relative alias, you must first understand the meaning of the cu
...
 
-  In this example, this is the page `/docs/grafana/latest/alerting/`.
+  In this example, this is the page `/docs/grafana/latest/`.
 
-The following table demonstrates the redirect URL, **REDIRECT**, created when using the relative alias, **RELATIVE ALIAS**, in the source file for **DESTINATION**:
+The following table demonstrates how to redirect **FROM PAGE**, to **TO PAGE**.
```
JStickler, line 107:
```suggestion
In the following table, the **FROM PAGE** column lists the page that requires a redirect, for example because the page has been moved or no longer exists. The **TO PAGE** is where we want to redirect them, for example the renamed page or where the content has been moved.
The third column shows the alias that should be added to the front matter of the file in the **TO PAGE** column to create the proper redirect.
```

JStickler (commented):
A couple of suggestions that follow up on how we talked about how this is similar to using `cd` in the meeting the other day.

docs/sources/write/front-matter/index.md
jdbaldry: I think we should avoid "easy" if possible (https://developers.google.com/style/word-list#easy).

My only concern with this tip is that it is only true at the start of the alias, and that's really because the starting directory for any alias is the directory containing the current page.

I'm worried people will internalize `.` goes back one part and `..` goes back two when in actually `.` goes nowhere and `..` goes back one part.

docs/sources/write/front-matter/index.md
JStickler: I don't think so.

docs/sources/write/front-matter/index.md
JStickler: Ah, good point about "easy".  

Describing how aliases work is hard!

docs/sources/write/front-matter/index.md
jdbaldry: I've attempted to rewrite this bit using a list of the columns and your suggestions for explaining each column in https://github.com/grafana/writers-toolkit/pull/396/commits/462fe219617fcc67146f0b876b4f5cc947dbfc9a

jdbaldry:
Gonna get this in to make sure the example is no longer broken but happy to accept additional contributions in follow up PRs.

PR #398, opened 2023-12-07
Add a couple of examples for convenience

osg-grafana:
When figuring out how to markup a heading when it is in a URL, it is helpful to include some examples so that it catches my eye. Perhaps this is also useful for others too.

### DIFF
--- a/docs/sources/write/links/index.md
+++ b/docs/sources/write/links/index.md
@@ -105,7 +105,7 @@ To convert a heading to an anchor, make the following changes:
 1. Convert to lower case.
 1. Remove any period characters (`.`).
 1. Replace any character that's not a lower cased letter, a number, or an underscore (`_`) with dashes (`-`).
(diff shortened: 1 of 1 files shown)


docs/sources/write/links/index.md
```diff
@@ -105,6 +105,7 @@ To convert a heading to an anchor, make the following changes:
 1. Convert to lower case.
 1. Remove any period characters (`.`).
 1. Replace any character that's not a lower cased letter, a number, or an underscore (`_`) with dashes (`-`).
+   Some examples include `.`, `/`, or `\`.
```
jdbaldry, line 108: Periods are removed by the previous step.

Should it be "and" rather than "or"?

```suggestion
   Some examples are `(`, `)`, `/`, and `\`.
```

I'll let others decide if this makes the step more understandable. To me it's redundant and potentially distracting but I'm not the primary consumer of these docs.

PR #404, opened 2023-12-14
Add cascade and param info to Front Matter and Shortcodes topics

clayton-cornell:
This PR adds `cascade` and `param` info to Front Matter and shortcodes topics in the Writer's Toolkit

### DIFF
--- a/docs/sources/write/shortcodes/index.md
+++ b/docs/sources/write/shortcodes/index.md
@@ -406,7 +406,7 @@ cascade:
 The `param` shortcode in the topic body text:
 
 ```markdown
-Welcome to Grafana {{</* param PRODUCT_VERSION */>}}! Read on to learn about changes to dashboards and visualizations, data sources, security and authentication, and more.
+Welcome to Grafana {{</* param PRODUCT_VERSION */>}}. Read on to learn about changes to dashboards and visualizations, data sources, security and authentication, and more.
 ```
 
 The `param` shortcode in a topic heading:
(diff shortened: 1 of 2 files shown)


docs/sources/write/shortcodes/index.md
```diff
@@ -383,6 +383,38 @@ In this example, the image's display size is changed to have a maximum width of
...
+
+```yaml
+cascade:
+  PRODUCT_VERSION: 10.2
+```
+
+The `param` shortcode in the topic body text:
+
+```markdown
+Welcome to Grafana {{</* param PRODUCT_VERSION */>}}! Read on to learn about changes to dashboards and visualizations, data sources, security and authentication, and more.
```
jdbaldry, line 409: I don't think we should have exclamations in examples since they are discouraged by the style guide.

jdbaldry (approved):
Nice! I like this a lot. The `cascade` documentation is a useful subset of the Hugo docs. There are alternative forms of cascading front matter that can selectively target a page but I think keeping to this single form is valuable because it keeps things simple.

docs/sources/write/shortcodes/index.md
clayton-cornell: I copied this direct from the published docs :-)  I'll remove it here.

clayton-cornell:
> alternative forms of cascading

I saw that when I was reading up on cascade. I see it's also used for labels. I only documented the variable... thinking we could expand on this section as we need.

PR #408, opened 2024-01-03
Docs: add directional language guidance

imatwawana:
Per a discussion in the style roundtable meeting.

### DIFF
--- a/docs/sources/write/style-guide/inclusive-writing/index.md
+++ b/docs/sources/write/style-guide/inclusive-writing/index.md
@@ -95,3 +95,12 @@ Use the following approach to describe relationships between nodes or processes:
 - Use _secondary_, _replica_, or _child_, instead of _slave_.
 
 Avoid _master_ or _slave_.
+
+### Right and left
+
+Using the terms _right_ and _left_ to denote a location is okay as long as you use them with a qualifier like _top_ or _bottom_ like so:
+
+- top-right
+- bottom-left
+
+Avoid using _right_ and _left_ without a qualifier.
--- a/docs/sources/write/style-guide/inclusive-writing/index.md
+++ b/docs/sources/write/style-guide/inclusive-writing/index.md
@@ -98,9 +98,11 @@ Avoid _master_ or _slave_.
 
 ### Right and left
 
-Using the terms _right_ and _left_ to denote a location is okay as long as you use them with a qualifier like _top_ or _bottom_ like so:
+Use the terms _right_ and _left_ with a qualifier like _top_ or _lower_ like so:
 
 - top-right
 - bottom-left
+- upper-left
+- lower-right
 
 Avoid using _right_ and _left_ without a qualifier.
--- a/docs/sources/write/style-guide/inclusive-writing/index.md
+++ b/docs/sources/write/style-guide/inclusive-writing/index.md
@@ -98,7 +98,7 @@ Avoid _master_ or _slave_.
 
 ### Right and left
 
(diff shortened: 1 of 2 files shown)


JStickler (approved):
LGTM

imatwawana:
@jdbaldry - I thought about adding this to Vale, but the words "right" and "left" seem like they'd come up for good reasons too often for this to be a good candidate.

docs/sources/write/style-guide/inclusive-writing/index.md
```diff
@@ -95,3 +95,14 @@ Use the following approach to describe relationships between nodes or processes:
 - Use _secondary_, _replica_, or _child_, instead of _slave_.
 
 Avoid _master_ or _slave_.
+
+### Right and left
```
josmperez, line 99: Do we want to say anything about use of the word "hand" as a qualifier, as in "top-right hand side"? Probably not, just want to be sure.

jdbaldry (commented):
Do we want to include the reasoning which comes from the Microsoft style, perhaps by linking to it or by including part of their justification:

> Don't use left as a directional term by itself. Instead, use a term such as upper left, lower left, leftmost, or on the left side of. Include a hyphen when you use left to modify a noun, as in upper-left corner. Don't use left-hand as a modifier.

> Accessibility tip Don't use directional terms (left, right, up, down) as the only clue to location. Individuals with cognitive disabilities might have difficulty interpreting them, as might people who are blind and use screen-reading software. It's OK to use a directional term if another indication of location, such as in the Save As dialog box, on the Standard toolbar, or in the title bar, is also included. Directional terms are also OK to use when a sighted user with dyslexia can clearly see a change in the interface as the result of an action, such as a change in the right pane when an option in the left pane is selected.

docs/sources/write/style-guide/inclusive-writing/index.md
imatwawana: It's already in the Vale linter since this is an easier rule to implement, but it probably doesn't hurt to add it.

docs/sources/write/style-guide/inclusive-writing/index.md
```diff
@@ -95,3 +95,14 @@ Use the following approach to describe relationships between nodes or processes:
...
+
+Use the terms _right_ and _left_ with a qualifier like _top_ or _lower_ like so:
+
+- top-right
+- bottom-left
+- upper-left
+- lower-right
+
+Avoid using _right_ and _left_ without a qualifier.
```
imatwawana, line 108: I think the last line about not using these words without a qualifier is redundant and messes with the flow if I add this bit about what qualifiers to not use. 
```suggestion

Don't use the word _hand_ as a qualifier (for example, _right-hand corner_).
```

imatwawana:
> > Don't use left as a directional term by itself. Instead, use a term such as upper left, lower left, leftmost, or on the left side of. Include a hyphen when you use left to modify a noun, as in upper-left corner. Don't use left-hand as a modifier.
> 
> > Accessibility tip Don't use directional terms (left, right, up, down) as the only clue to location. Individuals with cognitive disabilities might have difficulty interpreting them, as might people who are blind and use screen-reading software. It's OK to use a directional term if another indication of location, such as in the Save As dialog box, on the Standard toolbar, or in the title bar, is also included. Directional terms are also OK to use when a sighted user with dyslexia can clearly see a change in the interface as the result of an action, such as a change in the right pane when an option in the left pane is selected.

Can do!

josmperez (approved):

jdbaldry (approved):
<3

PR #419, opened 2024-01-22
Add heading exception for cloud traces content

knylander-grafana:
Adds "Reduce Grafana Cloud Traces costs" as an exception to the Headings list.


vale/Grafana/Headings.yml
```diff
@@ -34,3 +34,4 @@ exceptions:
   - TraceQL
   - What's new
   - YouTube
+  - Reduce Grafana Cloud Traces costs
```
jdbaldry, line 37: I think we just need the Grafana Cloud Traces bit as an exception

```suggestion
  - Grafana Cloud Traces
```

jdbaldry (approved):
Thanks for adding this!

PR #427, opened 2024-01-23
Note that `doc-validator` does not check full URL links

jdbaldry:
Also make clear that links can be to either versioned or unversioned documentation.

### DIFF
--- a/docs/sources/write/links/index.md
+++ b/docs/sources/write/links/index.md
@@ -38,8 +38,13 @@ For more information and examples, refer to [`docs/reference` shortcode](https:/
 
 ## Link to grafana.com pages
 
-Use a full URL with version substitution syntax (if needed).
+Use a full URL.
 
+{{< admonition type="note" >}}
+The `doc-validator` linter does not check links that use full URLs.
+{{< /admonition >}}
+
+If you are linking to versioned documentation, use a full URL with version substitution syntax.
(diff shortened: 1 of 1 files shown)


docs/sources/write/links/index.md
```diff
@@ -38,8 +38,13 @@ For more information and examples, refer to [`docs/reference` shortcode](https:/
...
 ## Link to grafana.com pages
 
-Use a full URL with version substitution syntax (if needed).
+Use a full URL.
 
+{{< admonition type="note" >}}
+The `doc-validator` linter does not check links that use full URLs.
+{{< /admonition >}}
+
+If you are linking to versioned documentation, use a full URL with version substitution syntax.
```
grafsean, line 47:
```suggestion
If you are linking to versioned documentation, use a full URL with version substitution syntax, without the protocol and hostname (https://grafana.com).
```

grafsean (approved):
Small change, feel free to edit it.

docs/sources/write/links/index.md
jdbaldry: You should see the exact same behavior with the protocol and hostname as without it.

We actually use the terminology "full URL" to indicate that those parts are allowed. It's easier to explain "full URLs" than the behavior of partial URLs.

PR #436, opened 2024-01-25
Add trailing slash guidance

imatwawana:


JStickler (approved):
Thank you, I can never remember this!

docs/sources/write/links/index.md
```diff
@@ -97,12 +97,12 @@ https://grafana.com/docs/mimir/<MIMIR_VERSION>/release-notes/
 
 ## Link to external pages
 
-Use the full URL.
+Use the full URL. Include trailing slashes.
```
jdbaldry, line 100: In this case, you just want to use the URL of the page as it is.
Trailing slashes are a feature of Hugo sites called "pretty URLs" and are not universal on the internet.

docs/sources/write/links/index.md
```diff
@@ -120,6 +120,7 @@ From a different page:
 ```markdown
 Read more in the [Grafana Open Source section of the Introduction page](https://grafana.com/docs/grafana/<GRAFANA_VERSION>/fundamentals/#grafana-open-source).
 ```
+Include the trailing slash before the hash (#) that precedes the page heading.
```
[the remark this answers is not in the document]
jdbaldry:
```suggestion
Include the trailing slash before the hash (#) that precedes the page heading for pages on the Grafana website.
```

docs/sources/write/links/index.md
```diff
@@ -97,12 +97,12 @@ https://grafana.com/docs/mimir/<MIMIR_VERSION>/release-notes/
...
 ## Link to external pages
 
-Use the full URL.
+Use the full URL. Include trailing slashes.
 
 For example:
 
 ```markdown
-https://github.com
+https://github.com/
```
jdbaldry, line 105:
```suggestion
https://github.com
```

knylander-grafana:
Thank you for adding this!

docs/sources/write/links/index.md
imatwawana: What about?
```suggestion
Use the full URL. Include trailing slashes where they are present. If a trailing slash isn't present when you copy the URL, don't add one.
```

imatwawana: @jdbaldry What if we had two examples for this? One with and one without a trailing slash? Also when I go to github.com and copy the full URL, it includes a trailing slash.

jdbaldry: The github.com one is interesting because it does indeed copy with a trailing slash.

I think the core of the advice is just copy the URL as it is.

If you go to https://github.com and copy the address bar you get `https://github.com/`
If you go to https://github.com/grafana/grafana and copy the address bar you get `https://github.com/grafana/grafana`.

imatwawana: Right, but the thing that people are unsure about is the trailing slash, so I want to speak to that.

imatwawana: What about
```suggestion
Use the full URL. Copy the URL as it is from the address bar. If it includes a trailing slash, include it; if it doesn't, don't.
```

jdbaldry (approved):
LGTM, thank you!

PR #440, opened 2024-01-30
Fix Vale rules for use of 'dialog box'

jdbaldry:
- Match 'dialog' only if it is not proceeded by ' box'.
- Link directly to the Grafana documentation heading for further information because it deviates from Google style.
- Explain the reason for the deviation.

### DIFF
--- a/docs/sources/write/style-guide/word-list/index.md
+++ b/docs/sources/write/style-guide/word-list/index.md
@@ -15,9 +15,9 @@ In most cases, you can refer to the [word list in the Google developer documenta
 - The word isn't included in Google guidelines.
 - It's still easy to use an incorrect word because it's widely used, generally or in other Grafana media.
 
-{{% admonition type="note" %}}
+{{< admonition type="note" >}}
 This page is a work in progress.
-{{% /admonition %}}
+{{< /admonition >}}
 
 <!-- vale off -->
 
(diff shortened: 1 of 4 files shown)


docs/sources/write/style-guide/word-list/index.md
```diff
@@ -15,9 +15,9 @@ In most cases, you can refer to the [word list in the Google developer documenta
 - The word isn't included in Google guidelines.
 - It's still easy to use an incorrect word because it's widely used, generally or in other Grafana media.
 
-{{% admonition type="note" %}}
+{{< admonition type="note" >}}
```
knylander-grafana, line 18: Should we be changing our admonition notes throughout our docs to use `{{<` instead of `{{%`?

jdbaldry: It's not urgent to make the change but it's really not necessary to use `{{%`, `{{<` is much more consistent with the behavior we want to see and means the shortcode is used the same as all others (except `docs/reference` where we actually need the behavior of `{{%`)

- https://raintank-corp.slack.com/archives/C067CLR6NVD/p1706331027643159

clayton-cornell: I just did a quick update in the Agent docs in this PR https://github.com/grafana/agent/pull/6279  It's fairly fast with some not-so-elegant find and sed commands.

clayton-cornell (approved):
LGTM

josmperez (commented):
LGTM.

PR #461, opened 2024-02-06
Lint 'Tooling and workflows' page and clean up duplicate documentation that I missed

jdbaldry:


docs/sources/write/tooling-and-workflows/index.md
```diff
@@ -89,18 +92,18 @@ From github.com:grafana/mimir
...
 ```
 
-### Create a new branch from the main remote branch
+### Create a branch from the default remote branch
 
 By convention, the remote repository in GitHub is the source of truth for a repository's history.
-The main branch of a repository is typically called `main` and occasionally called `master`.
-Note that we prefer to use inclusive language, so `main` is the preferred name.
+The default branch of a repository is typically called `main` and occasionally called `master`.
+Note that Grafana prefers to use inclusive language, so `main` is the preferred name.
```
clayton-cornell, line 99: Should we format this as a Note vs "Note that...."?  Or Just state "Grafana prefers to use...."

docs/sources/write/tooling-and-workflows/index.md
```diff
@@ -301,7 +318,8 @@ git switch main
 git merge --ff-only
 ```
 
-After switching branches, git also lets you know if your local copy needs updating ("Your branch is behind 'origin/main'..."). The `--ff-only` flag means "fast-forwarding", which simply applies the changes to the end of your local copy because your local copy has no conflicting changes.
+After switching branches, git also lets you know if your local copy needs updating with the message _Your branch is behind 'origin/main'_.
```
clayton-cornell, line 321:
```suggestion
After switching branches, Git also lets you know if your local copy needs updating with the message _Your branch is behind 'origin/main'_.
```

Should we be referring to Git with initial caps? For reference: https://git-scm.com/ uses Git instead of git.

I only tagged this one instance... it's lowercase git throughout this topic (and others).

clayton-cornell (approved):
minor nits :-)

docs/sources/write/tooling-and-workflows/index.md
jdbaldry: Good call out. I think if it is _Git_ in their documentation, it should be _Git_ in ours. I'll add a Vale rule.

docs/sources/write/tooling-and-workflows/index.md
jdbaldry: I don't think it justifies a full note so:

```suggestion
Grafana prefers to use inclusive language, so `main` is the preferred name.
```

Issue #455, opened 2024-02-02
Understand sentence case heading rule and explain it better in the bot feedback


jdbaldry:
For example, the heading `### Install Grafana Agent in static mode using Ansible` fails the rule.

I believe it is to do with the threshold for non-capitalized words since Grafana Agent and Ansible are both in the sentence case exceptions.

I need to confirm this understanding and also explain it in the error message so people can understand a false positive if it occurs.

PR #474, opened 2024-02-13
Add test fixtures for `Grafana/Heading` rule that demonstrates threshold behavior

jdbaldry:
Closes https://github.com/grafana/writers-toolkit/issues/455


vale/Grafana/Headings.yml
```diff
@@ -2,7 +2,12 @@ extends: capitalization
 message: |
   Use sentence-style capitalization for '%s'.
 
-  If your heading uses sentence capitalization but this error is reported, you need to add an exception to https://github.com/grafana/writers-toolkit/blob/main/vale/Grafana/Headings.yml.
+  If your heading contains capitalized words that represent product names, you need to add those words as exceptions in https://github.com/grafana/writers-toolkit/blob/main/vale/Grafana/Headings.yml for them to be considered correctly cased.
+
+  The threshold for sentence casing is 0.3.
+  That means that the ratio of correctly cased words to all words in the heading must be above 0.3.
+
+  Vale considers multi-word exceptions such as _Grafana Enterprise Metrics_ as a single correctly cased word in threshold calculations.
```
heitortsergent, line 10: I'm wondering if this is helpful for reviewers, I had to read it a few times to understand it. Would this help explain in case someone sees a heading that should trigger a warning and that's not the case?

jdbaldry: It's mostly to explain the behavior in:

https://github.com/grafana/writers-toolkit/pull/474/files#diff-db77cd56bbd111e2274758441944ce965a8157bb4204b359daacc4801f4c77d3R44-R48

Depending on how you understand the rule you could believe that three of the words are correctly sentence cased (Grafana, Enterprise, and Metrics), but Vale only considers one word to be correctly sentence cased (Grafana Enterprise Metrics as a whole).

jdbaldry: Do you think it would be better to omit this in the message and avoid the overload of information?

clayton-cornell: Is it the 0.3 part that's confusing here? Is it really necessary to state the threshold at this point in the topic?  Could we simplify lines 7 - 10 down to this:

```suggestion
  Vale considers multi-word exceptions such as _Grafana Enterprise Metrics_ as a single correctly cased word.
```

jdbaldry: I think the threshold is definitely a confusing aspect. I think people understand when a product name is causing the alert to fire but find it confusing when the rule is triggered by the threshold. Perhaps that's gonna be less confusing now that I've tuned down the threshold.

I'm down to remove the extra explanation and keep an eye on the rule.

jdbaldry:
Good question. It's a convention I've borrowed from the RedHat Vale repo (example https://github.com/redhat-documentation/vale-at-red-hat/tree/main/.vale/fixtures/RedHat/Contractions).

The Markdown file has a single example of a working heading and so the golden file is empty (Vale outputs nothing when run on the Markdown input).

clayton-cornell (approved):
Looks OK to me. All makes sense now ;-)

PR #486, opened 2024-02-13
docs: adds terms to wordlist

chri2547:
This PR adds "KPI" and "KPIs" to the wordlist so that they won't be flagged by the linter.

### DIFF
--- a/vale/dictionaries/en_US-grafana.dic
+++ b/vale/dictionaries/en_US-grafana.dic
@@ -1,4 +1,4 @@
-81
+83
 Aerospike/ po:noun
 Alertmanager/S po:noun
 allowlist/S po:noun
@@ -36,6 +36,7 @@ JMeter/ po:noun
 Jsonnet/ po:noun
 JUnit/ po:noun
 Kinesis/ po:noun
+KPI/S po:noun
 kubelet/ po:noun
 misconfiguration/S po:noun
 namespace/S po:noun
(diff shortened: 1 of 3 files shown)


jdbaldry (commented):
Thank you for the contribution! Is this abbreviation also worth excluding from the https://github.com/grafana/writers-toolkit/blob/main/vale/Grafana/Acronyms.yml rule?

My leaning is that we should probably still explain this term on its first use in each page.

vale/dictionaries/en_US-grafana.dic
```diff
@@ -36,6 +36,8 @@ JMeter/ po:noun
 Jsonnet/ po:noun
 JUnit/ po:noun
 Kinesis/ po:noun
+KPI/ po:noun
+KPIs/ po:noun
```
jdbaldry, line 40: This is just a nit of how the dictionary works but with a `/S` we can have a single entry for this work plus the pluralization which comes from the affix rule in https://github.com/grafana/writers-toolkit/blob/main/vale/dictionaries/en_US-grafana.aff#L20-L24.

```suggestion
KPI/S po:noun
```

chri2547: @jdbaldry oh cool, thanks!

jdbaldry: I've already applied the suggestion but I've left the comment unresolved as explanation.

PR #491, opened 2024-02-14
Review _Build and review documentation_ page

jdbaldry:
Closes https://github.com/grafana/writers-toolkit/issues/489

### DIFF
--- a/docs/sources/review/_index.md
+++ b/docs/sources/review/_index.md
@@ -1,19 +1,21 @@
 ---
-title: Build and review documentation
-menuTitle: Build and review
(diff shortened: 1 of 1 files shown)


clayton-cornell (approved):
The first time I've noticed or seen the date metadata.

jdbaldry:
Yeah for RSS feeds and stuff that uses the `date` front matter in the website build, we tend to rely on the value Hugo pulls out of the Git log. 

Since I'm lazy, my tool for notifying me of documentation that needs review only checks the front matter explicitly and doesn't try to derive the date of the last modification from Git.

PR #493, opened 2024-02-14
Add matcher to dictionary

jdbaldry:
It is a Prometheus term for the labels specified in a query used to match specific series.


clayton-cornell (approved):
LGTM

PR #496, opened 2024-02-15
Add spanset to dictionary

jdbaldry:
Defined in https://github.com/grafana/website/pull/18319

### DIFF
--- a/vale/dictionaries/en_US-grafana.dic
+++ b/vale/dictionaries/en_US-grafana.dic
@@ -1,4 +1,4 @@
-85
+86
 Aerospike/ po:noun
 Alertmanager/S po:noun
 allowlist/S po:noun
@@ -67,6 +67,7 @@ showback/S po:noun
 siloed/ po:adjective
 SLI/S po:noun
 SLO/S po:noun
+spanset/S po:noun
 Splunk/ po:adjective
 Splunk/ po:noun
 SRE/S po:noun
--- a/vale/dictionaries/en_US-grafana.wordlist
+++ b/vale/dictionaries/en_US-grafana.wordlist
@@ -66,6 +66,7 @@ showback/S po:noun
 siloed/ po:adjective
 SLI/S po:noun
 SLO/S po:noun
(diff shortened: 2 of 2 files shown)


vale/dictionaries/en_US-grafana.dic
```diff
@@ -1,4 +1,4 @@
-85
+86
```
clayton-cornell, line 1: I'm interested in this value. I've tried poking the Vale docs, but not sure what to search for. I can see it's a count of entries in this file.. but for what purpose? What happens if we forget to increment this value?

clayton-cornell (approved):

vale/dictionaries/en_US-grafana.dic
jdbaldry: Ah yeah, this is not actually a Vale thing but rather an aspect of the Hunspell dictionary format. 

The best resource I have right now is https://manpages.ubuntu.com/manpages/focal/man5/hunspell.5.html.

The need to count the entries is actually why I have a separate `.wordlist` file that we can just add entries to and then run `make all` which generates the `.dic` file from the `.wordlist` file so the number is always right.

knylander-grafana (approved):
Thank you for adding this term, Jack!

jdbaldry:
What do you think about adding it to the [glossary](https://github.com/grafana/website/blob/master/data/glossary.yaml?rgh-link-date=2024-02-15T10%3A08%3A22Z)?

It would be useful for this PR to be able to reference that as justification for the word being in the dictionary.

knylander-grafana:
> What do you think about adding it to the [glossary](https://github.com/grafana/website/blob/master/data/glossary.yaml?rgh-link-date=2024-02-15T17%3A33%3A53Z)?
> 
> It would be useful for this PR to be able to reference that as justification for the word being in the dictionary.

Sure. We don't currently have spanset created as an entry. I'll see if we have an existing definition some where. j

jdbaldry:
Awesome, thank you!

PR #501, opened 2024-02-15
Review "Links" documentation

jdbaldry:
In its previous form, readers were confused into believing the only behavior of version substitution was front matter lookup which provides little value over linking without the syntax.

The rewrite intends to more clearly state that the default behavior is inference from the current page but that it can be overridden with front matter.

### DIFF
--- a/docs/sources/write/links/_index.md
+++ b/docs/sources/write/links/_index.md
@@ -1,11 +1,10 @@
 ---
-title: Links
-description: Understand how to link between pages.
-weight: 600
 aliases:
   - /docs/writers-toolkit/write/links/
   - /docs/writers-toolkit/write/references/
   - /docs/writers-toolkit/writing-guide/references/
+date: 2024-02-15
+description: Understand how to link between pages.
 keywords:
   - Hugo
   - link
@@ -14,6 +13,8 @@ keywords:
   - ref
   - references
   - relref
+title: Links
+weight: 600
 ---
 
 # Links
@@ -21,7 +22,7 @@ keywords:
 Choose your link type based on the applicable scenario:
 
 - [Link from source content that's used in multiple projects](#link-from-source-content-thats-used-in-multiple-projects)
-- [Link to grafana.com pages](#link-to-grafanacom-pages)
+- [Link to `grafana.com` pages](#link-to-grafanacom-pages)
 - [Link to external pages](#link-to-external-pages)
 - [Link to page headings](#link-to-page-headings)
 
@@ -36,20 +37,24 @@ Use the `docs/reference` shortcode.
 The source is reused as described in [Reuse directories of content with Hugo mounts](https://grafana.com/docs/writers-toolkit/write/reuse-content/reuse-directories/).
 For more information and examples, refer to [`docs/reference` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#docsreference).
 
-## Link to grafana.com pages
+## Link to `grafana.com` pages
 
 Use a full URL.
 
 {{< admonition type="note" >}}
-The `doc-validator` linter does not check links that use full URLs.
+The `doc-validator` linter doesn't check links that use full URLs.
 {{< /admonition >}}
 
-If you are linking to versioned documentation, use a full URL with version substitution syntax.
-Version substitution is necessary for full URLs to link to the correct version of documentation.
-Usually, this is the current version of documentation.
+If you are linking to versioned documentation, use a full URL with version substitution syntax instead of the version path element.
+For example, in Grafana, use `<GRAFANA_VERSION>` instead of `latest` in the URL `https://grafana.com/docs/grafana/latest/`.
 
-In versioned documentation, set the correct version in the root `_index.md` file for your documentation.
-The following YAML example merges with the existing front matter in the root `_index.md` file, and sets `GRAFANA_VERSION` to be `latest` for that page and all child pages.
+Version substitution replaces the `<SOMETHING_VERSION>` syntax with the version inferred from the current page.
+It's necessary for full URLs to link to the correct version of documentation without requiring the writer to update the version for each release.
+For examples of behavior, refer to [Examples](#examples).
+
+To override the version inferred by version substitution, set the preferred version in the root `_index.md` file for your documentation.
+The following YAML snippet sets `GRAFANA_VERSION` to be `latest` for that page and all child pages.
+You must merge the following YAML example with the front matter in the root `_index.md` file.
 
 ```yaml
 cascade:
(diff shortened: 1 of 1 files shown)


docs/sources/write/links/_index.md
```diff
@@ -36,20 +37,24 @@ Use the `docs/reference` shortcode.
...
 
-If you are linking to versioned documentation, use a full URL with version substitution syntax.
-Version substitution is necessary for full URLs to link to the correct version of documentation.
-Usually, this is the current version of documentation.
+If you are linking to versioned documentation, use a full URL with version substitution syntax instead of the version path element.
+For example, in Grafana, use `<GRAFANA_VERSION>` instead of `latest` in the URL `https://grafana.com/docs/grafana/latest/`.
 
-In versioned documentation, set the correct version in the root `_index.md` file for your documentation.
-The following YAML example merges with the existing front matter in the root `_index.md` file, and sets `GRAFANA_VERSION` to be `latest` for that page and all child pages.
+Version substitution replaces the `<SOMETHING_VERSION>` syntax with the version inferred from the current page.
```
JStickler, line 51:
```suggestion
When the documentation is published to the website, version substitution replaces the `<SOMETHING_VERSION>` syntax with the version inferred from the current branch for the documentation page.
```

docs/sources/write/links/_index.md
```diff
@@ -36,20 +37,24 @@ Use the `docs/reference` shortcode.
...
+It's necessary for full URLs to link to the correct version of documentation without requiring the writer to update the version for each release.
```
JStickler, line 52:
```suggestion
You use full URLs with the substitution variable so that links will resolve to the correct version of documentation without requiring the writer to update the version for each release.
```

JStickler (commented):
Thanks for taking the time to revise this. I was going to take a crack at it this morning, but you beat me to it. That said, I have a couple of suggestions for how to make this clearer.

docs/sources/write/links/_index.md
jdbaldry: It's not really true that version substitution happens during publishing. It happens when the link is rendered. Perhaps:

```suggestion
When Hugo renders links with version substitution, it replaces the `<SOMETHING_VERSION>` syntax with the version inferred from the current page.
```

docs/sources/write/links/_index.md
```diff
@@ -36,20 +37,24 @@ Use the `docs/reference` shortcode.
...
 {{< admonition type="note" >}}
-The `doc-validator` linter does not check links that use full URLs.
+The `doc-validator` linter doesn't check links that use full URLs.
 {{< /admonition >}}
 
-If you are linking to versioned documentation, use a full URL with version substitution syntax.
-Version substitution is necessary for full URLs to link to the correct version of documentation.
-Usually, this is the current version of documentation.
+If you are linking to versioned documentation, use a full URL with version substitution syntax instead of the version path element.
+For example, in Grafana, use `<GRAFANA_VERSION>` instead of `latest` in the URL `https://grafana.com/docs/grafana/latest/`.
```
clayton-cornell, line 49: This kind of caught me out the first time I read it.. .I read it as just use `<GRAFANA_VERSION>` as the URL (despite knowing the correct way).  I don't have a suggestion here... just noting that it's one of those "read carefully" sentences.

clayton-cornell (approved):
Looks OK to me. One minor observation.. no suggestions though.

docs/sources/write/links/_index.md
```diff
@@ -36,20 +37,24 @@ Use the `docs/reference` shortcode.
...
 
-In versioned documentation, set the correct version in the root `_index.md` file for your documentation.
-The following YAML example merges with the existing front matter in the root `_index.md` file, and sets `GRAFANA_VERSION` to be `latest` for that page and all child pages.
+When Hugo renders links with version substitution, it replaces the `<SOMETHING_VERSION>` syntax with the version inferred from the current page.
```
JStickler, line 51: As far as I know, there is zero data in a markdown topic to indicate the version.  If I'm understanding correctly how this  magic works, isn't it the BRANCH that the page is being published to that determines the version?

clayton-cornell: Agent uses this feature to link to specific versions of external and internal documentation... and sets the referenced version as a `cascade`.

jdbaldry: It comes from the relative permalink for the page rather than from the source file. Hugo knows the relative permalink for the page at build time and that contains a version that we extract as part of the link.

The page `docs/sources/loki/_index.md` from the Loki repository is synced to the `content/docs/loki/<VERSION>/` directory in the website repository and at Hugo build time the page has relative permalink `/docs/loki/<VERSION>/`.

It is from that relative permalink that we can extract the version and is what is intended by the phrase "inferred from the page".

If you think in terms of pages rather than source files, it perhaps is simpler.
If you are on the page `/docs/grafana/latest/dashboards/` version substitution replaces the `<GRAFANA_VERSION>` in links with `latest` from the page URL.

All of this is the default behavior. Front matter overrides this if it is set.

JStickler: Could we add this information to the Writers Toolkit?  Because without understanding what magic Hugo is working behind the scenes, the way this is currently written implies that the version is somehow extracted from the markdown file for a page.  Which doesn't seem possible (and isn't, as you've described what's really happening).

jdbaldry: I think that's in https://grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution. What about linking to that?

Initially the advice from Isabel for this page was to focus on what to do rather than how stuff works.
Unfortunately, it feels to me that depending on who looks at the page, we endlessly oscillate back and forth from explaining everything to just telling people what to do.

```suggestion
When Hugo renders links with version substitution, it replaces the `<SOMETHING_VERSION>` syntax with the version inferred from the current page.

To understand the behavior in more detail, refer to [About version substitution](https://grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution)
```

JStickler: > we endlessly oscillate back and forth from explaining everything to just telling people what to do.

I admit, I am the Elephant's Child, always asking "Why?"  I don't know that the About Version Substitution page makes any more sense to me.  

Maybe we should just revise this line to read "When Hugo renders links with version substitution, it replaces the `<SOMETHING_VERSION>` syntax with the correct version."  and not say anything about how the magic happens.

clayton-cornell: I'm a big supporter of NOT mentioning the magic behind the curtain in the user/contributor facing documentation. Save the details for a child page or a "how it all works" doc set for the maintainers. Here, just saying "it does a thing when you do this" is probably sufficient... and either link out to the how it works, or keep the how it works in a design doc set somewhere else?

docs/sources/write/links/_index.md
jdbaldry: Mm, thanks for raising this. I think it is more clear in https://grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution:~:text=As%20a%20convention%2C%20use%20the%20name%20of%20the%20target%20project%20all%20upper%2Dcase.%20For%20example%2C%20grafana%20becomes%20GRAFANA%2C%20grafana%2Dcloud%20becomes%20GRAFANA%20CLOUD.

Perhaps that should be here and the "About ..." should more on the behavior under the hood.

PR #506, opened 2024-02-18
Add rule for catching use of blockquotes instead of admonitions

jdbaldry:
I'm gonna use it so I can fix up all the blockquotes in Mimir.

This will cause false positives on repositories that rely on `docs/reference` so I probably won't merge this until https://github.com/grafana/website/pull/18304 has replaced `docs/reference`.


clayton-cornell (approved):
LGTM

PR #512, opened 2024-02-21
Add separate rule for Kubernetes capitalization

jdbaldry:


clayton-cornell:
Is this the right place to define more K8s-specific terminology?  For example:
* Kubectl
* Kubelet
* Minikube
* etc.

jdbaldry:
Good question. This is a substitution rule so we can definitely add some substitutions here :)

- Kubectl: I think we should suggest only "`kubectl`" (https://kubernetes.io/docs/contribute/style/style-guide/#use-code-style-for-kubernetes-command-tool-and-component-names)
- Kubelet: similarly, only "`kubelet`" (https://kubernetes.io/docs/contribute/style/style-guide/#use-code-style-for-kubernetes-command-tool-and-component-names)
- Minikube: it looks like the preferred name in the project is "minikube" (https://minikube.sigs.k8s.io/docs/) which would need to also be added to the dictionary so the spellchecker knows of it.

clayton-cornell:
Oooops, they got capitalized in my list :-)  I'm not so sure we use `minikube` in our docs, but within Agent, we use `kubectl` and `kubelet` a lot.

clayton-cornell:
Will these fail if they are lowercase in a component name?  Example: `k8s.daemonset.name` is an attribute in the `otelcol.processor.k8sattributes` component.

jdbaldry:
Not in `inline code`, or **bold**, we've configured Vale to ignore those tokens because of https://grafana.com/docs/writers-toolkit/write/style-guide/style-conventions/#bold

vale/fixtures/Grafana/Kubernetes/testinvalid.md
```diff
@@ -0,0 +1,10 @@
+Don't use:
+
+- cron job
+- deamonset
+- Kubernetes deployment
+- pod
+- replicaset
+- statefulset
+- kubelet
+- kubectl
```
clayton-cornell, line 10: Do we need tests for Kubectl and Kubelet (with initial caps)?

jdbaldry: Good catch, added in [2f9a383](https://github.com/grafana/writers-toolkit/pull/512/commits/2f9a3839edf1bc75362eb3e03774ad1d1293e0d2)

clayton-cornell (approved):
This looks great to me now :-)

PR #536, opened 2024-02-23
Docs: add SQL to word list

imatwawana:
While it's not controversial that _SQL_ is pronounced _sequel_ when you're talking about Microsoft SQL Server, it's less clear how to pronounce the acronym—and thus whether to use the article "a" or "an"—in other contexts. 

Per our [style roundtable discussion](https://docs.google.com/document/d/1uCjtHEy0lpZggkpgsLDjWwxV9K6h7vcsG2dy4TOS6EA/edit#heading=h.1vctzxeg938f), we've decided to follow IBM's guidance on which article to use with the acronym "SQL" because Structured Query Language originated at IBM. Some [examples](https://www.ibm.com/docs/en/i/7.5?topic=ldap-troubleshooting-directory-server):

- “When _an SQL error_ occurs, the QDIRSRV job log will usually contain the following message:..."
- "However, in some cases QDIRSRV might not contain this message and this referral, even if _an SQL server_ is the cause of the problem.”

Notably, this does not align with guidance from [Microsoft](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/s/sql-sql-server) or [Google](https://developers.google.com/style/articles).

**NOTE**: When we had the discussion, we were using a reference from an older MS Manual of Style and we didn't check their current style guide. At the time, we thought MS was aligned with IBM.


JStickler (approved):

jdbaldry (approved):
LGTM!

josmperez (approved):
I disagree, but agree to approve and move on!

chri2547 (approved):

PR #537, opened 2024-02-25
Add Vale rule for SQL usage

jdbaldry:
The fixtures Markdown files demonstrate valid and invalid use and the related `.golden` files show the Vale output from those Markdown files.

It adds a special case for the ambiguous case of "a SQL server".
I can add more test cases to those files if you would like to try out other words.

The `[Aa] SQL server` regular expression can be read as: match the word `a`, possibly capitalized, followed by the string of words ` SQL server`.

The `[Aa]n SQL Server` regular expression can be read as: match the word `an`, possibly capitalized, followed by the string of words ` SQL Server`.

The `[Aa] SQL(?! [Ss]erver)` regular expression can be read as: match the word `a`, possibly capitalized, followed by the word ` SQL`, only if it is NOT followed by a space and the word `server`, possibly capitalized.


imatwawana (approved):
LGTM!

PR #541, opened 2024-02-26
Review "About our documentation design" page

jdbaldry:
Closes https://github.com/grafana/writers-toolkit/issues/538.

- Justifies use of first person
- Use correct emphasis for phrases in sentences.

### DIFF
--- a/docs/sources/structure/about-documentation-design/index.md
+++ b/docs/sources/structure/about-documentation-design/index.md
@@ -1,29 +1,39 @@
 ---
-title: About our documentation design
-menuTitle: Documentation design
-description: Learn about Grafana's documentation docs pages
-weight: 100
 aliases:
   - /docs/writers-toolkit/writing-guide/about-documentation-design/
   - /docs/writers-toolkit/structure/about-documentation-design/
+date: 2024-02-26
+description: Learn about the design of Grafana's documentation pages
 keywords:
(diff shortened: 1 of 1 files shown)


docs/sources/structure/about-documentation-design/index.md
```diff
@@ -1,29 +1,39 @@
...
   - Grafana
   - documentation
   - page design
+menuTitle: Documentation design
+title: About documentation design
+weight: 100
 ---
 
-# About our documentation design
+# About documentation design
```
JStickler, line 16: Not sure why you removed "our" throughout this topic?  This is not a generic topic, it's specifically about **_Grafana Labs_** documentation design.

jdbaldry: Good question.  I guess we can revise it to start with "About documentation design at Grafana Labs" or something similar before we and an exception for the use of personal pronouns for the rest of the page and then we would be following the style guidance in https://developers.google.com/style/pronouns#personal-pronouns.

jdbaldry: I agree, exceptions are very important. I will revise this to start with a clear reference to our organization and then add the exception comment.

JStickler (approved):
LGTM

PR #571, opened 2024-03-04
Generate rewrites

jdbaldry:
The idea is that this single file will replace the aliases throughout all of the pages according to the proposal in [Easy to use redirects for documentation pages](https://docs.google.com/document/d/1aGpLQDXBeD5wOvR-qtaRds7uqZbArF1B1YH3i4MhkWc/edit#heading=h.r5tgv62q8045).

The idea is that you can simply write the absolute URLs in future (with `<PROJECT_VERSION>` instead of "latest" for versioned docs and the website will handle the logic to make these redirects behave properly across all versions.

For example, to redirect https://grafana.com/docs/grafana/latest/datasources/ to https://grafana.com/docs/grafana/latest/data-sources/ (with the correct hypen):

```
/docs/grafana/<GRAFANA_VERSION>/datasources/ /docs/grafana/<GRAFANA_VERSION>/data-sources/
```


chri2547 (approved):

jdbaldry:
> I can't think of any gotchas on this. I like the abstraction... the only "hard part" is remembering to add the alias here when someone else (devs) make structural changes. In the current process, it's a little more in front of them with the aliases metadata (although, I don't think many really understand how it works in the current implementation).
> 
> tl;dr LGTM

Yeah that's a fair concern. We could write a little script that checks that: if there are renamed files, that there are also changes to the rewrites file. We could even post that as a comment on the PR or file.

clayton-cornell:
> a little script that checks that: if there are renamed files, that there are also changes to the rewrites file

That would be very useful for everyone... a little nudge that reminds us isn't a bad thing (I know i need reminders).

PR #574, opened 2024-03-05
Add "VUH" to wordlist

heitortsergent:
"VUH" is an acronym for Virtual User hours that we use regularly in the k6 docs.

### DIFF
--- a/vale/dictionaries/en_US-grafana.dic
+++ b/vale/dictionaries/en_US-grafana.dic
@@ -1,4 +1,4 @@
-96
+97
 Aerospike/ po:noun
 Alertmanager/S po:noun
 allowlist/S po:noun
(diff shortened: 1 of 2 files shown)


clayton-cornell:
I wonder if maybe we should be using `VUH` instead of `VUh` here?

A quick Google on the acronym returns results where everyone buy Grafana seems to be using `VUH`. For example:
* https://azure.microsoft.com/en-ca/pricing/details/load-testing/
* https://guides.flood.io/pricing/what-is-a-vuh
* https://help.mabl.com/hc/en-us/articles/19078248250900-Performance-testing-FAQs
and many more.

heitortsergent:
@clayton-cornell that's a good point. I'll ask the k6 team and see if they have any strong feelings about using "VUh" over "VUH".

jdbaldry (approved):
LGTM

PR #575, opened 2024-03-06
Add flamegraph to en_US-grafana.wordlist

knylander-grafana:
Added flamegraph to the word list for Vale.

### DIFF
--- a/vale/dictionaries/en_US-grafana.wordlist
+++ b/vale/dictionaries/en_US-grafana.wordlist
@@ -25,6 +25,7 @@ etcd/ po:noun
 ESLint/ po:noun
 Fargate/ po:noun
 Firehose/ po:noun
+Flamegraph/ po:noun
 Goldmark/ po:noun
 Grafana/ po:adjective
 Grafana/M po:noun


jdbaldry (approved):
I think flamegraph is jargon and is typically written as one word. However, the author of the flamegraph, Brendan Gregg, writes it in two words: https://www.brendangregg.com/flamegraphs.html

This is a really interesting one because the existing suffix rule would expect "flamegraphes" because anything ending with an "h" is suffixed with "es" https://github.com/grafana/writers-toolkit/pull/575/files#diff-a866c302e6905a20dcba419725819c9f4a1d02502f1b782971825bff55483dceL23.

I've added an additional case for "ph" -> "phs". Let me know if you know of a significant number of exceptions to that rule.

jdbaldry:
> > I think flamegraph is jargon and is typically written as one word. However, the author of the flamegraph, Brendan Gregg, writes it in two words: [brendangregg.com/flamegraphs.html](https://www.brendangregg.com/flamegraphs.html)
> 
> I asked about whether this is one word or two words in the Pyroscope dev channel. The response was the same as you mentioned above: we consistently use it as one word but the original author uses it as two. The earliest content on Pyroscope.io uses flamegraph as one word as well. However, to be consistent with all of the UI and existing doc for Pyroscope, I'd vote for keeping flamegraph as one word. We can always revisit it later.
> 
> > This is a really interesting one because the existing suffix rule would expect "flamegraphes" because anything ending with an "h" is suffixed with "es" [#575 (files)](https://github.com/grafana/writers-toolkit/pull/575/files#diff-a866c302e6905a20dcba419725819c9f4a1d02502f1b782971825bff55483dceL23).
> 
> That is interesting. Maybe it isn't the singular "h" but the grouping of letters with the "H". The word "graphes" looks wrong to me, while "graphs" looks correct. Although, not everything ending in an "h" uses "es" as a suffix. For example, words ending in "-th" add an s (month > months), while words ending in -sh or -ch add an -es (marsh > marshes). See [en.wikipedia.org/wiki/English_plurals#Irregular_plurals](https://en.wikipedia.org/wiki/English_plurals#Irregular_plurals) for some additional examples.
> 
> > I've added an additional case for "ph" -> "phs". Let me know if you know of a significant number of exceptions to that rule.
> 
> This would be an excellent question for the style roundtable discussion, honestly. I know there are others besides gh, ph, th, sh, ch. We'll need to dig out some heavy style guides and grammar rules to find all of the answers.

❤️ great analysis. I think this suffix rule should be fine for now but I'll definitely bring the topic up if it will get people interested :)

clayton-cornell:
> anything ending with an "h" is suffixed with "es

That seems... wrong. It should be anything ending in `s`, `sh`, `ch`, `x`, or `z`, you would add an `es`... not just `h`.

PR #577, opened 2024-03-07
Add "boolean" and "CIDR" to the words list

mariomac:


jdbaldry (commented):
According to Wikipedia, the data type is a Boolean (https://en.wikipedia.org/wiki/Boolean_data_type) and if we refer to types in code that are called `boolean`, the inline code excludes the word from spell checking.

That being said, when I type boolean in my browser, it doesn't flag it as a spelling error and I'm happy to be pragmatic and add this in.

mariomac:
Hi @jdbaldry! I added boolean here because I got an error when mentioning "boolean" inside a table:
https://github.com/grafana/beyla/pull/675/files#diff-4a8c4879d91a378852e53817dbbcc43c838145645cc4e38d53160d74e10bcf7aR42

clayton-cornell:
This is a topic that generates LOTS of discussion. I mumbled into this early in my work on Agent. The agreement at that time was that:
1. You use initial caps when it's an adjective: Boolean logic
2. You don't use initial caps when it's a noun: boolean as a data type.

So this PR makes sense and LGTM :-)

clayton-cornell (approved):

jdbaldry:
That sounds like great advice. I'll add boolean as a noun. `po:<THING>` refers to the part of speech so in this case `po:noun` is correct and not `po:adjective`.

vale/dictionaries/en_US-grafana.wordlist
```diff
@@ -8,9 +8,11 @@ autoscale/DGRS po:verb
 backport/DGS po:verb
 Beyla/ po:noun
 blockquote/S po:noun
+boolean/ po:adjective
 BPF/ po:noun
 burndown/ po:adjective
 cAdvisor/ po:noun
+CIDR/s po:noun
 CRD/s po:noun
```
jdbaldry, line 16:
```suggestion
CIDR/S po:noun
CRD/S po:noun
```

vale/dictionaries/en_US-grafana.wordlist
```diff
@@ -8,9 +8,11 @@ autoscale/DGRS po:verb
 backport/DGS po:verb
 Beyla/ po:noun
 blockquote/S po:noun
+boolean/ po:adjective
```
jdbaldry, line 11:
```suggestion
boolean/S po:noun
```

jdbaldry (commented):
Updating the entries

PR #580, opened 2024-03-11
Add some Mimir words to the dictionary

jdbaldry:


jdbaldry (commented):
There are more to add but I only had so much time and energy for that today...

clayton-cornell (approved):
Looks OK to me. I didn't double check every term, but.. nothing jumps out at me as odd :-)

PR #588, opened 2024-03-16
Remove flamegraph in favor of flame graph

knylander-grafana:
Remove flamegraph. The content team chimed in and decided we should use flame graph as two words. I've reverted the change I made earlier and have also reviewed all of the content in the Cloud Profiles content to make flame graph two words in the documentation. I'll update the Pyroscope OSS content next. 

Part of https://github.com/grafana/pyroscope-squad/issues/123


clayton-cornell (approved):

jdbaldry (approved):
Thanks for checking this and coming back to revise the dictionary :)

PR #589, opened 2024-03-18
Document `ref` URIs that replace `docs/reference` shortcodes

jdbaldry:
You can use `ref` URIs in admonitions (or any shortcodes) because they are inline and not subject to the issues noted in the [`admonition` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#code-shortcode:~:text=to%20core%20understanding.-,WARNING,For%20more%20information%2C%20refer%20to%20Markdown%20Reference%20Links%20in%20Shortcodes.,-Examples). 

The `ref` URIs perform the same pattern matching as `docs/reference` but don't require the use of reference-style links and the destinations are ordinary (full) URLs that can include version substitution. 
Unlike `docs/reference`, the implementation doesn't use `relref` so you don't have to be careful with omitting trailing slashes and the links will follow redirects.

I have requested review from anyone that I know is having to use `docs/reference` in their documentation.
Please review this documentation and ensure that it makes sense to you so that, when I open a PR to replace `docs/reference` with `ref` URIs, you can maintain these links going forward.

I am happy to answer any questions and address an design concerns before this is merged, I'll make the appropriate changes to the implementation if needed.


brendamuir (approved):

docs/sources/write/shortcodes/index.md
```diff
@@ -409,8 +409,14 @@ Produces:
 
 ## Docs/reference
 
+{{< admonition type="warning" >}}
+This shortcode is present in the documentation, but you should prefer `ref` URIs.
+Don’t use it when creating new or updating existing documentation.
+
```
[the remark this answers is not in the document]
jdbaldry: I kept the copy as close to https://grafana.com/docs/writers-toolkit/write/shortcodes/#relref for consistency. Do you think both should be changed. I think your suggestion has a subtly different meaning.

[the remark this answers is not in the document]
brendamuir: its super nit picky so dont feel you have to change it. just read better to me that way--but thats just me:)

docs/sources/write/links/_index.md
```diff
@@ -19,23 +19,40 @@ weight: 600
...
+```markdown
+[LINK TEXT](ref:<KEY>)
+```
+
+Hugo looks up _`KEY`_ in the value for the `refs` field in the page's front matter.
+If there is no _`KEY`_ in the `refs` field, or there is no `refs` field in the front matter, Hugo logs a build error.
 
-The source is reused as described in [Reuse directories of content with Hugo mounts](https://grafana.com/docs/writers-toolkit/write/reuse-content/reuse-directories/).
-For more information and examples, refer to [`docs/reference` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#docsreference).
+{{< docs/shared source="writers-toolkit" lookup="refs-example.md" leveloffset="+2" >}}
```
imatwawana, line 55: Given that this example text will be here, it seems like lines 44-53 would be redundant. I'll have to build to check, but just wanted to get this review in before this got merged.

imatwawana (commented):
Thanks for this update! Big win! I'll circle back once I've built this to see if my last comment stands.

docs/sources/shared/refs-example.md
```diff
@@ -0,0 +1,49 @@
...
+
+In the [Grafana page](https://grafana.com/docs/grafana/latest/alerting/set-up/), the link destination is https://grafana.com/docs/grafana/latest/alerting/fundamentals/data-source-alerting/.
+Hugo replaces the version substitution syntax `<GRAFANA_VERSION>` with the version inferred from the current page.
+For more information about version substitution, refer to [About version substitution](https://grafana.com/docs/writers-toolkit/write/shortcodes/#about-version-substitution).
+
+In the page [Grafana Cloud page](https://grafana.com/docs/grafana/latest/alerting/set-up/), the link destination is https://grafana.com/docs/grafana-cloud/alerting-and-irm/alerting/fundamentals/alert-rules/.
+
+```markdown
+---
+refs:
```
imatwawana, line 36: Just a general question: does this mean that the references that used to live at the end of many pages will now live in the front matter?

jdbaldry: Indeed :)

docs/sources/write/links/_index.md
```diff
@@ -19,23 +19,40 @@ weight: 600
...
-- [Link to page headings](#link-to-page-headings)
 
-Although these other types of links still function, replace them with full URLs:
+You may want to link to a heading within a page.
+To do so, use one of the preceding options and then refer to [Link to page headings](#link-to-page-headings).
+
+Although these other types of links still function, replace them with one of the preceding options:
 
 - [Hugo `relref` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#relref)
+- [`docs/reference shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#docsreference)
```
imatwawana, line 34:
```suggestion
- [`docs/reference` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#docsreference)
```

docs/sources/write/links/_index.md
```diff
@@ -19,23 +19,40 @@ weight: 600
...
+Choose your link type based on your goal:
 
 - [Link from source content that's used in multiple projects](#link-from-source-content-thats-used-in-multiple-projects)
 - [Link to `grafana.com` pages](#link-to-grafanacom-pages)
 - [Link to external pages](#link-to-external-pages)
-- [Link to page headings](#link-to-page-headings)
 
-Although these other types of links still function, replace them with full URLs:
+You may want to link to a heading within a page.
+To do so, use one of the preceding options and then refer to [Link to page headings](#link-to-page-headings).
```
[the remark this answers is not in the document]
jdbaldry: Cool that makes sense :)

docs/sources/shared/refs-example.md
```diff
@@ -0,0 +1,49 @@
...
+  grafana-alerting:
```
imatwawana, line 37: It seems like the key has to be formatted as all one word? I'm assuming that based on the hyphen in the example key. If not, then maybe don't include a hyphen in this one. If so, then we should state that explicitly.

docs/sources/write/links/_index.md
imatwawana: Okay I see what you're trying to do with this section. In preview, it's a hard read, so lemme think on how we might reformat it a bit and circle back.

docs/sources/shared/refs-example.md
jdbaldry: Yeah, let's get that written down, the full restrictions are technically different, but saying that it needs to be one dash-separated or underscore-separated word is easy

imatwawana:
Just an overall comment, I think we need to socialize this a bit before we pull the trigger. Happy to help on that front.

jdbaldry:
> Just an overall comment, I think we need to socialize this a bit before we pull the trigger. Happy to help on that front.

Before the docs are live or before we make the changes in the source repos?

imatwawana:
> > Just an overall comment, I think we need to socialize this a bit before we pull the trigger. Happy to help on that front.
> 
> Before the docs are live or before we make the changes in the source repos?

The latter. Maybe once we update these docs, it would be good to do some work to make sure all the reviewers listed here are super clear on things. And then we can maybe make an announcement in #docs-grafana or something.

docs/sources/shared/refs-example.md
imatwawana: Love the updated functionality; hate scrolling past all that front matter, lol. I'll get used to it.

clayton-cornell (approved):

docs/sources/shared/refs-example.md
```diff
@@ -0,0 +1,49 @@
...
+Hugo checks each `pattern` and `destination` pair, in order, against the page's URL.
+If the page's URL matches the pattern _`URL PATH PREFIX`_, then Hugo uses `destination` as the link destination.
+If no pattern matches the current page, Hugo logs a build error.
+
+The _`FULL URL`_ destination has the same behavior as described in [Link to `grafana.com` pages](https://grafana.com/docs/writers-toolkit/write/links/#link-to-grafanacom-pages)
+
+# Example
+
+The following Markdown snippet demonstrates the `refs` front matter and link that uses a `ref` URI.
+
```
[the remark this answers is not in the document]
jdbaldry: Fixed in c622a6f

docs/sources/shared/refs-example.md
jdbaldry: Fixed in 749627a

docs/sources/write/links/_index.md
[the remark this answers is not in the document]
jdbaldry: Fixed in c4d75fc

chri2547 (approved):

jdbaldry:
@imatwawana I think this is ready for a second pass when you have time.

clayton-cornell:
I'm still a bit unclear on these ref URIs in the metadata section. Are we suggesting that ALL xrefs in a topic be defined here? Or only those that need special handling because of differing destinations due to different mounts and paths? What about xref links within the doc tree?  Eg... a link that would be defined as a relative URI like this: `[link text]: ../../topic/`.  This stays within the local doc tree and doesn't have a different path/target in any mount point.

jdbaldry:
> I'm still a bit unclear on these ref URIs in the metadata section. Are we suggesting that ALL xrefs in a topic be defined here? Or only those that need special handling because of differing destinations due to different mounts and paths? What about xref links within the doc tree? Eg... a link that would be defined as a relative URI like this: `[link text]: ../../topic/`. This stays within the local doc tree and doesn't have a different path/target in any mount point.

I would personally suggest that you use whatever is easiest for you. We don't have any specific guidance around relative URLs because they are another thing to explain. We found that with `relref` the relative part is more complicated than the concept of full URLs for some or most technical writers.

The idea of this documentation is that it's a simple decision tree, if you are writing mounted content, use ref URIs.

If I were writing my own documentation set and I didn't have to worry about explaining it to all technical writers, I would do as you proposed and use relative URLs for all the internal links and only use `ref` URIs when I was linking to two places where the relative path is not the same for both.

docs/sources/write/links/_index.md
```diff
@@ -19,23 +19,43 @@ weight: 600
...
 
 ## Link from source content that's used in multiple projects
 
-Use the `docs/reference` shortcode.
+Use `ref` URIs.
 
-The source is reused as described in [Reuse directories of content with Hugo mounts](https://grafana.com/docs/writers-toolkit/write/reuse-content/reuse-directories/).
-For more information and examples, refer to [`docs/reference` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#docsreference).
+{{< admonition type="note" >}}
+If you are using `doc-validator` in CI, you must upgrade to v5.0.0 to use `ref` URIs.
```
imatwawana, line 40:
```suggestion
If you're using `doc-validator` in CI, you must upgrade to v5.0.0 to use `ref` URIs.
```

docs/sources/write/links/_index.md
```diff
@@ -19,23 +19,43 @@ weight: 600
...
+
+`doc-validator` no longer checks link destinations.
+You must manually check link destinations in the local preview or fix broken links identified by the weekly website crawl.
+{{< /admonition >}}
+
+`ref` URIs lookup destinations based upon the page's URL path and the definitions in the page's front matter.
```
imatwawana, line 46:
```suggestion
`ref` URIs look up destinations based upon the page's URL path and the definitions in the page's front matter.
```

docs/sources/shared/refs-example.md
```diff
@@ -0,0 +1,48 @@
...
+  <KEY>:
+    - pattern: <URL PATH PREFIX>
+      destination: <FULL URL>
+    - pattern: <URL PATH PREFIX>
+      destination: <FULL URL>
+```
+
+Hugo checks each `pattern` and `destination` pair, in order, against the page's URL.
+If the page's URL matches the pattern _`URL PATH PREFIX`_, then Hugo uses `destination` as the link destination.
+If no pattern matches the current page, Hugo logs a build error.
```
imatwawana, line 20: Can this be less of an explanation of Hugo behaviour and more a piece of direction. I've left in a bit of tht content, but I wonder if we can even just take that out. Maybe something like (but obviously this isn't exactly right)?
```suggestion

- _`URL PATH PREFIX`_ - Enter the part of the page URL that represents the project in which these docs will appear. 
- `destination` - Enter the full URL of the destination page for that project including trailing slashes.

If the page's URL matches the pattern _`URL PATH PREFIX`_, then Hugo uses `destination` as the link destination.
If no pattern matches the current page, Hugo logs a build error.
```

docs/sources/write/links/_index.md
```diff
@@ -19,23 +19,43 @@ weight: 600
...
+A link with a `ref` URI looks like:
+
+```markdown
+[LINK TEXT](ref:<KEY>)
+```
+
+Hugo looks up _`KEY`_ in the value for the `refs` field in the page's front matter.
+_`KEY`_ must only use alphabetic and hyphen (`-`) characters.
+If there is no _`KEY`_ in the `refs` field, or there is no `refs` field in the front matter, Hugo logs a build error.
+
```
imatwawana, line 57: Same feedback as the corresponding section in the example shared file:
```suggestion

- _`KEY`_ - Enter an alphabetic key. It can include hyphens (`-`).

Hugo looks up _`KEY`_ in the value for the `refs` field in the page's front matter.
If there is no _`KEY`_ in the `refs` field, or there is no `refs` field in the front matter, Hugo logs a build error.

```

imatwawana (commented):
I've made some suggestions for tweaking the formatting and approach for the link text and front matter explanations that might make it easier for folks to understand what they need to do vs what the system is going to do with what they add.

docs/sources/write/links/_index.md
```diff
@@ -19,23 +19,43 @@ weight: 600
...
-Although these other types of links still function, replace them with full URLs:
+Although these other types of links still function, replace them with one of the preceding options:
 
 - [Hugo `relref` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#relref)
+- [`docs/reference` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#docsreference)
 
 ## Link from source content that's used in multiple projects
 
-Use the `docs/reference` shortcode.
+Use `ref` URIs.
```
imatwawana, line 37:
```suggestion
Use `ref` URIs. `Ref` URIs have two components:

- [Link](#link)
- [Front matter](#front-matter)
```

docs/sources/write/links/_index.md
```diff
@@ -19,23 +19,43 @@ weight: 600
...
-For more information and examples, refer to [`docs/reference` shortcode](https://grafana.com/docs/writers-toolkit/write/shortcodes/#docsreference).
+{{< admonition type="note" >}}
+If you are using `doc-validator` in CI, you must upgrade to v5.0.0 to use `ref` URIs.
+
+`doc-validator` no longer checks link destinations.
+You must manually check link destinations in the local preview or fix broken links identified by the weekly website crawl.
+{{< /admonition >}}
+
+`ref` URIs lookup destinations based upon the page's URL path and the definitions in the page's front matter.
+
```
imatwawana, line 47:
```suggestion

### Link

```

docs/sources/write/links/_index.md
```diff
@@ -19,23 +19,43 @@ weight: 600
...
+```markdown
+[LINK TEXT](ref:<KEY>)
+```
+
+Hugo looks up _`KEY`_ in the value for the `refs` field in the page's front matter.
+_`KEY`_ must only use alphabetic and hyphen (`-`) characters.
+If there is no _`KEY`_ in the `refs` field, or there is no `refs` field in the front matter, Hugo logs a build error.
+
+{{< docs/shared source="writers-toolkit" lookup="refs-example.md" leveloffset="+2" >}}
```
imatwawana, line 58:
```suggestion
### Front matter

{{< docs/shared source="writers-toolkit" lookup="refs-example.md" leveloffset="+2" >}}
```

imatwawana (approved):

PR #594, opened 2024-03-20
Docs: add guidance for second entries

imatwawana:
Adds guidance for when you need to announce that a previously documented feature has moved into a new release stage.

### DIFF
--- a/docs/sources/contribute/release-notes/index.md
+++ b/docs/sources/contribute/release-notes/index.md
@@ -180,6 +180,12 @@ To understand the process of creating release notes for Grafana versioned releas
 
 If you add an entry to the CMS after the relevant versioned What's new has already been published, you'll need to open a PR to also add it to the versioned What's new yourself.
 
+#### Updated release stage
+
+If you've previously created a What's new entry for a feature when it was in an early release stage and you want to announce that the feature has moved into a new release stage, create a new entry. For example, you published a What's new entry when the feature was in public preview and now the feature is in general availability.
+
+You don't have to make the new entry as robust as the previous one. Instead, you can keep the new entry brief by referring to or linking to refer the previous entry.
+
 ### Edit What's new entries
 
 Regardless of the status of your entry, it's always best to use the CMS to make any changes.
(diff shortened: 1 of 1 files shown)


jdbaldry (approved):
Nice LGTM.

Usual nit about semantic line breaks: 
[Semantic line breaks](https://sembr.org/) are better for future line based diffing in the GitHub UI and result in the same HTML output.

docs/sources/contribute/release-notes/index.md
```diff
@@ -180,6 +180,12 @@ To understand the process of creating release notes for Grafana versioned releas
 
 If you add an entry to the CMS after the relevant versioned What's new has already been published, you'll need to open a PR to also add it to the versioned What's new yourself.
 
+#### Updated release stage
+
+If you've previously created a What's new entry for a feature when it was in an early release stage and you want to announce that the feature has moved into a new release stage, create a new entry. For example, you published a What's new entry when the feature was in public preview and now the feature is in general availability.
+
+You don't have to make the new entry as robust as the previous one. Instead, you can keep the new entry brief by referring to or linking to refer the previous entry.
```
JStickler, line 187:
```suggestion
You don't have to make the new entry as robust as the previous one. Instead, you can keep the new entry brief by referring to or linking to the previous entry.
```

JStickler (approved):
Just one tiny tweak, otherwise LGTM

imatwawana:
> Nice LGTM.
> 
> Usual nit about semantic line breaks: [Semantic line breaks](https://sembr.org/) are better for future line based diffing in the GitHub UI and result in the same HTML output.

For me personally, semantic line breaks tend to impede my comprehension of the content rather than improve it if I'm reading it in the editor vs in a local preview. So even if I was going to add them, I'd have to do all my writing and then go back and add them. Given that we haven't decided to require them, it's just more overhead for me to add them.

jdbaldry:
Different priorities I suppose. I spend most of my time reviewing diff patches and in local previews so semantic line breaks are more important to me. 

I don't personally find it any harder to read two sentences over two lines instead of two sentences on one line but perhaps that's primarily being exposed to reading code line by line and I'm reading the docs like code rather than like prose, I'm not sure.

clayton-cornell:
I'm on the side of liking semantic line breaks. Some of the dev team I work with do 80 column breaks and it ends up being a mess to review/edit - especially once suggestions are applied. Semantic line breaks really simplify the review edit cycle.... for me.

I don't think the shoudl be required, but as an option for readability? For me it works.

imatwawana:
We've already landed at a "not required" conclusion on this. Ultimately folks are free to use them if they want and free to not use them if they don't want to.

PR #598, opened 2024-03-25
Adding entry for Gravatar

melgl:
This adds an entry for [Gravatar](https://gravatar.com/), the tool used for populating user avatars in Grafana instances.


clayton-cornell (approved):
LGTM

PR #601, opened 2024-04-04
fixed GRAFANA_VERSION variable

usmangt:
in a few places, it is written as GRAFANA VERSION which is not correct


jdbaldry (approved):
Thanks bud!

PR #612, opened 2024-04-10
Docs: add ordinal number guidance

imatwawana:
Adds a Numbers section to the style guide. Adds guidance to follow Google style on all number styling except in the case of ordinals.

### DIFF
--- a/docs/sources/write/style-guide/style-conventions/index.md
+++ b/docs/sources/write/style-guide/style-conventions/index.md
@@ -173,6 +173,12 @@ For more information about Grafana Labs products, refer to [Grafana documentatio
 
 For more comprehensive guidance, refer to [Write useful link text](https://grafana.com/docs/writers-toolkit/write/links/useful-link-text/)
 
+## Numbers
+
+For direction on how to style numbers, follow the [Google style guide](https://developers.google.com/style/numbers) except in the case of ordinals.
+
+For ordinals, write out first through ninth. For 10th on, use numerals.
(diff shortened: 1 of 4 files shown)


clayton-cornell (approved):

docs/sources/write/style-guide/style-conventions/index.md
```diff
@@ -173,6 +173,12 @@ For more information about Grafana Labs products, refer to [Grafana documentatio
 
 For more comprehensive guidance, refer to [Write useful link text](https://grafana.com/docs/writers-toolkit/write/links/useful-link-text/)
 
+## Numbers
+
+For direction on how to style numbers, follow the [Google style guide](https://developers.google.com/style/numbers) except in the case of ordinals.
```
josmperez, line 178:
```suggestion
For direction on how to style numbers, follow the [Google style guide](https://developers.google.com/style/numbers) except in the case of _ordinals_. An ordinal number is a number that indicates the position or order of something in relation to other numbers, like, first, second, third, and so on.
```

josmperez:
I have added a suggest to define ordinal, since non-writers may be reading this and be unfamiliar with the term. Arguably, it was already defined in context, but I think this suggestion is a bit clearer.

jdbaldry:
I've opened https://github.com/grafana/writers-toolkit/pull/615 to have Vale enforce the Writers' Toolkit guidance and not the Google guidance.

PR #615, opened 2024-04-11
Update ordinal rule to match Writers' Toolkit guidance 

jdbaldry:


imatwawana (approved):
Looks great!

PR #618, opened 2024-04-12
Generate acronym exceptions from dictionary

jdbaldry:
Removes:
  - ASP which doesn't seem to be an established acronym in our area
  - XSS which I think benefits from being explained


clayton-cornell (approved):
Looks OK to me

PR #619, opened 2024-04-12
Clean up some word list behavior

jdbaldry:
- Ensure traceql regexp is parsed correctly.
- Fix false positives for Kinesis matchers that were not mutually exclusive with the substitution.
- Remove Kubernetes rules which exist in Grafana/Kubernetes.yml

Also run `prettier` on the Vale rules


clayton-cornell (approved):
Simple change. LGTM :-)

PR #621, opened 2024-04-15
Review "Write useful link text" page

jdbaldry:
Closes https://github.com/grafana/writers-toolkit/issues/570


docs/sources/write/links/useful-link-text.md
```diff
@@ -15,7 +16,7 @@ The primary purpose of link text is to give users a short summary of the link de
 
 ## Summary
 
-When writing link text, pay attention to words, quantity of links, and the length of link text.
+When writing link text, pay attention to the words used and the length of link text.
```
jdbaldry, line 19: I don't think the quantity of links is important in writing link text.

chri2547, line 19:
```suggestion
When writing link text, pay attention to the words used and the length of the link text.
```

chri2547: Agreed. I wonder why that's there?

chri2547 (approved):
LGTM! Thanks for the update!

docs/sources/write/links/useful-link-text.md
jdbaldry: I think this may have previously been part of another page which was more generally about link text but I haven't actually confirmed that.

JStickler: I think the general intent with mentioning "quality" was probably to keep people from inserting too many lnks?  I have seen pages on the internet that have so many links that they're almost unreadable.

jdbaldry: Yeah I think that is likely. There is https://github.com/grafana/writers-toolkit/pull/621/files#diff-33fd5d831920c6dc2aa12d37245edb938412193e3309533ba11952d76e50e7bdL34-L40 which talks about that.

PR #628, opened 2024-04-16
Add overbill to dictionary

heitortsergent:
Related to https://github.com/grafana/website/pull/19354.


vale/dictionary.jsonnet
```diff
@@ -121,6 +121,7 @@ local newWord(word, affixes, po) = {
     newWord('OSS', '', 'noun') { acronym: true, established_acronym: true },
     newWord('OTel', '', 'adjective'),
     newWord('OTel', 'M', 'noun'),
+    newWord('overbill', '', 'verb'),
```
heitortsergent, line 124: I'm not sure about the second parameter here, do I need to add any values to it? 🤓

jdbaldry (approved):
Nice, thank you for the contribution!

vale/dictionary.jsonnet
heitortsergent: That's perfect, thanks @jdbaldry!

It might be nice to add that as a comment at the top of this file, for which letters to use and their meaning. The en_US-grafana.aff is a little bit difficult to parse. 😱

PR #634, opened 2024-04-17
Add some video guidelines

jdbaldry:


imatwawana (commented):
I think this is enough to start with. We can always add to this as we discover more use cases.

chri2547 (approved):
Thanks for adding these guidelines!

PR #639, opened 2024-04-24
Start documenting guidance for `docs/shared` usage

jdbaldry:


docs/sources/write/shortcodes/index.md
```diff
@@ -514,13 +514,18 @@ For more detailed instructions, refer to [Reuse shared content](https://grafana.
...
+| `leveloffset` | Manipulates source content headings up to a maximum level of `h6`. Only positive offsets are supported. `leveloffset="+5"` ensures an `h1` in the source content is an `h6` in the destination content.                                                                                                                        | no       |
 
 {{< admonition type="note" >}}
 Hugo doesn't rebuild the destination file when a source file changes on disk.
 To trigger a rebuild after changes to a source file, perform a trivial change to the destination file and save that, too.
 {{< /admonition >}}
 
+### Guidance
+
+When the page with the `docs/shared` shortcode includes a shared page from the same project, you should use version substitution syntax for the `version` parameter.
```
clayton-cornell, line 526: What do you do if it doesn't?

jdbaldry: If it doesn't, you should use the version of the docs that you want to include from. Usually I guess this would be latest. 

Do you think it would be useful to include that here too?

clayton-cornell: My question was prompted by the conditions in the sentence... if there is a shared page in the same project then use version substitution.  OK, so then what happens if there's a shared page that's not in the same project?

I think it'd be useful to cover both scenarios... otherwise we leave the reader hanging, wondering what to do in the "else" case.

Does this apply to say... a topic in /agent/shared that's reused in/shared to Loki? You don't use AGENT_VERSION anymore? You use LATEST?

docs/sources/write/shortcodes/index.md
```diff
@@ -514,13 +514,18 @@ For more detailed instructions, refer to [Reuse shared content](https://grafana.
...
+This ensures that the `docs/shared` shortcode includes the page from the same version as the page the shortcode is in.
```
jdbaldry, line 527:
```suggestion
This ensures that the `docs/shared` shortcode includes the page from the same version as the page the shortcode is in.

Otherwise, you should use a version appropriate for your documentation.
Use version substitution syntax and set the desired version in cascading front matter in your project's root `_index.md` file.
```

docs/sources/write/shortcodes/index.md
jdbaldry: To simplify the guidance, I've chosen to focus only on version substitution syntax in https://github.com/grafana/writers-toolkit/pull/639/files#r1602869550

docs/sources/write/shortcodes/index.md
clayton-cornell: This clarifies it :-)

PR #642, opened 2024-04-26
Add boolean to dictionary.jsonnet

heitortsergent:
Looking at the Google style guide and in the context of technical docs, I think we should include the lowercase boolean to our dictionary: https://developers.google.com/style/word-list#boolean.


clayton-cornell (approved):
Agreed from my side anyway :-) I use lowercase boolean in the Agent/Alloy docs

jdbaldry (approved):
Makes sense to me! Thanks for contributing and linking to the Google word list documentation.

PR #645, opened 2024-05-01
Address false positives in https://github.com/grafana/website/pull/19613/

jdbaldry:
[they renamed the service not long ago just to Data Firehose](grafana/website#19613 (comment))


clayton-cornell (approved):
Looks good. 

Just for reference: https://aws.amazon.com/firehose/ where you can see how the product name is used

PR #647, opened 2024-05-03
Update dictionary with Amazon RDS and generate AmazonProductNames and Headings rules

jdbaldry:


clayton-cornell (approved):
Looks good. Nothing jumped out at me :-)

PR #649, opened 2024-05-03
Note that first person plural is fine when referring to Grafana Labs

jdbaldry:
It's certainly preferable to passive voice!


vale/Grafana/We.yml
```diff
@@ -0,0 +1,16 @@
+extends: existence
+message: |
+  Use first person plural pronouns like '%s' carefully.
+
+  Don't use 'we' when you are talking about the reader, instead use 'you'.
+
+  It's OK to use 'we' when you are talking about Grafana Labs.
```
imatwawana, line 7:
```suggestion
  Don't use 'we' when you're talking about the reader, instead use 'you'.

  It's OK to use 'we' when you're talking about Grafana Labs.
```

PR #663, opened 2024-05-21
adding words for Faro source map upload docs

eskirk:
for use in https://github.com/grafana/website/pull/19808


jdbaldry (commented):
Thank you for the contribution :) 

Would you be able to add a description to both Rollup and Vite. I'm guessing they are https://rollupjs.org/ and https://vitejs.dev/

jdbaldry (approved):
Manually updated with descriptions because I'm confident in because of the context.

PR #686, opened 2024-05-29
Add retryable adjective

jdbaldry:
https://github.com/grafana/website/pull/20022/


vale/dictionaries/en_US-grafana.dic
```diff
@@ -1,4 +1,4 @@
-208
+210
```
clayton-cornell, line 1: Is this a number that we should update with each new entry?

clayton-cornell (approved):
LGTM

vale/dictionaries/en_US-grafana.dic
jdbaldry: Yeah, it should match the number of entries in the dictionary file.
I guess it's so tools that parse the dictionary can know the number of entries up front.

Fortunately it's generated each time from the dictionary.jsonnet so it should always be in sync with the count in the file.

It went up by two this time because I replaced one word (incorrect adject 'query') with another (correct adjective 'queryable') and then added entries for the verb 'retry' and adjective 'retryable'.

PR #700, opened 2024-05-31
Add Vale rule to lint for writing specific shortcodes with percent syntax

jdbaldry:
In most cases, it isn't needed.
For consistency and pattern matching, we should prefer {{< because it better reflects the shortcode semantics we default to in our minds.

The other syntax is only necessary in specific contexts and the nuances are complicated.

### DIFF
--- a/vale/Grafana/Shortcodes.yml
+++ b/vale/Grafana/Shortcodes.yml
@@ -1,0 +1,18 @@
+extends: script
+message: |
+  Use `{{<` and `>}}` instead of `{{%%` and `%%}}`.
+
+  The former has the most consistent semantics.
+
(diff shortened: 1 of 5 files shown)


vale/Grafana/Shortcodes.yml
```diff
@@ -0,0 +1,18 @@
+extends: script
+message: |
+  Use `{{<` and `>}}` instead of `{{%%` and `%%}}`.
```
jdbaldry, line 3: Note the %% is intentional because this is a format string and %s is used to substitute in values. In order to have a literal % you have to escape it with %%.

clayton-cornell (approved):

PR #705, opened 2024-05-31
Add overutilization to dictionary.jsonnet

heitortsergent:


clayton-cornell (approved):

vale/dictionary.jsonnet
```diff
@@ -143,6 +143,7 @@ local newWord(word, affixes, po) = {
     newWord('OTel', 'M', 'noun'),
     newWord('OTLP', '', 'noun') { acronym: true, established_acronym: true },
     newWord('overbill', 'DG', 'verb'),
+    newWord('overutilization', 'S', 'noun'),
```
jdbaldry, line 146: Is overutilizations a word? The 'S' affix is for pluralization.

clayton-cornell: I didn't really come up with much use of the plural of overutilization (web search)... although Miriam Webster does define it with the "s" as plural.

jdbaldry: True! Wild haha

Issue #669, opened 2024-05-22
Document how to contribute to the Vale dictionary


jdbaldry:
Cover the important parts of the Jsonnet and how the rules and dictionary are regenerated

derek-cadzow:
@jdbaldry Question from our triage this week -- status on this?  Is it something you're planning to assign yourself to?

jdbaldry:
Yeah, let me get this written this week so I can close this.

As a maintainer of this repository I also consider any open issue on this repository implicitly assigned to me in some senses.
I've explicitly assigned myself to this now that I'm planning to work on it, hopefully that helps with your triage :)

PR #708, opened 2024-06-03
Update dictionary.jsonnet

BeverlyJaneJ:
Hey, Jack/Clayton, trying to help out and add Wildfly. Don't know if I did it correctly.


jdbaldry:
I'm writing the doc for https://github.com/grafana/writers-toolkit/issues/669 today so I'll make sure to include that.

clayton-cornell (approved):

vale/dictionary.jsonnet
```diff
@@ -233,6 +233,7 @@ local newWord(word, affixes, po) = {
     newWord('walkthrough', 'S', 'noun'),
     newWord('Webpack', '', 'noun'),
     newWord('webserver', 'S', 'noun'),
+    newWord('Wildfly', '', 'noun'),
```
jdbaldry, line 236: It looks like it's actually WildFly, according to Wikipedia. Is that correct?

```suggestion
    newWord('WildFly', '', 'noun') { description: 'WildFly, formerly known as JBoss AS, or simply JBoss, is an application server written by JBoss, now developed by Red Hat (https://en.wikipedia.org/wiki/WildFly)', product: true },
```

clayton-cornell: I wonder if it's worth doing a refactor on the dictionary to add in the missing metadata (descriptions etc)...

PR #720, opened 2024-06-06
Add `review_date` front matter documentation

jdbaldry:


docs/sources/write/front-matter/index.md
```diff
@@ -298,6 +298,16 @@ Use the `refs` front matter with `ref` URIs to link to different pages in reused
 
 For more information, refer to [Link from source content that’s used in multiple projects](https://grafana.com/docs/writers-toolkit/write/links/#link-from-source-content-thats-used-in-multiple-projects).
 
+### Review date
+
+Use the `review_date` front matter to set the time you last reviewed a page for correctness.
```
chri2547, line 303:
```suggestion
Use the `review_date` front matter to set the date you last reviewed a page for correctness.
```

chri2547: Changing this to "date" because time is more specific, ie 2:13pm.

chri2547 (commented):
one nit, thanks!

docs/sources/write/front-matter/index.md
jdbaldry: Good point! I think you can actually set a very specific time but I omitted that for simplicity and given that omission, it's more correct to talk about dates and not times.

chri2547 (approved):

clayton-cornell (approved):

PR #721, opened 2024-06-07
Add rule for catching empty alt text in Markdown images

jdbaldry:


clayton-cornell (approved):
The example alt text is funny :-)

PR #725, opened 2024-06-07
Use correct placeholder syntax

jdbaldry:
I had misremembered the guidance in
https://grafana.com/docs/writers-toolkit/write/style-guide/write-for-developers/#placeholder-variables


clayton-cornell (approved):
I forget to check for this in the PRs.... :-P so you're not the only one who misremembers things. :-)

PR #727, opened 2024-06-10
Note that menuTitles should include the task topic verb

jdbaldry:


chri2547 (approved):

docs/sources/write/front-matter/index.md
```diff
@@ -288,7 +288,11 @@ cascade:
 
 <!-- vale Grafana.Headings = YES -->
 
-Use `menuTitle` to specify a different heading in the sidebar navigation than the `title` element; for example, if you want to abbreviate the topic heading in the table of contents.
+Use `menuTitle` to specify a different heading in the sidebar navigation than the `title` element.
```
clayton-cornell, line 291:
```suggestion
Use `menuTitle` to specify a different title in the sidebar navigation than in the `title` element.
```

docs/sources/write/front-matter/index.md
```diff
@@ -288,7 +288,11 @@ cascade:
...
+For example, if you want to abbreviate the topic heading in the table of contents.
+
+Don't remove the verb from [task topic](https://grafana.com/docs/writers-toolkit/structure/topic-types/task/) headings.
```
clayton-cornell, line 294: I'm not sure why this was included here. It seems pretty tangential to the `menuTitle` element.

jdbaldry: This is actually the crux of the change so I think I need to reword it. When deriving the menuTitle from the title, you must keep the verb to indicate that the page is a task before the reader clicks on it. I think the use of heading is making this more confusing?

docs/sources/write/front-matter/index.md
```diff
@@ -288,7 +288,11 @@ cascade:
 
 <!-- vale Grafana.Headings = YES -->
 
-Use `menuTitle` to specify a different heading in the sidebar navigation than the `title` element; for example, if you want to abbreviate the topic heading in the table of contents.
+Use `menuTitle` to specify a different heading in the sidebar navigation than the `title` element.
+For example, if you want to abbreviate the topic heading in the table of contents.
```
jdbaldry, line 292:
```suggestion
For example, if you want to abbreviate the topic title in the table of contents.
```

docs/sources/write/front-matter/index.md
jdbaldry, line 294:
```suggestion
Don't remove the verb from [task topic](https://grafana.com/docs/writers-toolkit/structure/topic-types/task/) titles.
```

jdbaldry: I've made the change to use 'title' everywhere. Does it make more sense now?

docs/sources/write/front-matter/index.md
```diff
@@ -288,7 +288,11 @@ cascade:
 
 <!-- vale Grafana.Headings = YES -->
 
-Use `menuTitle` to specify a different heading in the sidebar navigation than the `title` element; for example, if you want to abbreviate the topic heading in the table of contents.
+Use `menuTitle` to specify a different title in the sidebar navigation than in the `title` element.
+For example, if you want to abbreviate the topic title in the table of contents.
+
+Don't remove the verb from [task topic](https://grafana.com/docs/writers-toolkit/structure/topic-types/task/) titles.
+The verb helps the reader know that they're navigating to a task topic before they follow the link.
```
clayton-cornell, line 295:
```suggestion
Don't remove the verb from the `title`.
The verb helps the reader know that they're navigating to a  [task topic](https://grafana.com/docs/writers-toolkit/structure/topic-types/task/) before they follow the link.
```

What if we tweaked it a little more here. We want to keep the verb in the title metadata, correct? Which also then matches the H1 content.

jdbaldry: The verb should be in the `title` (and the first heading should match the `title`) AND it should also be in the `menuTitle` so that readers navigating via the table of contents menu know that they're navigating to a task topic before they follow the link.

jdbaldry, line 295:
```suggestion
Don't remove the verb from [task topic](https://grafana.com/docs/writers-toolkit/structure/topic-types/task/) titles.
The verb helps the reader know that they're navigating to a task topic before they follow the link.

It is OK to remove the verb from the `menuTitle` if it's implied by the containing section.
For example, the page 'Install Grafana Alloy in a Docker container' can be shortened to 'Docker' if it's in a section named 'Install'.
```

clayton-cornell (approved):
Looks good with clarifications as discussed :)

PR #728, opened 2024-06-10
Add contentful to dictionary.jsonnet

heitortsergent:
Related to https://github.com/grafana/website/pull/20024, I forgot to add this one. 🤓 

Word is related to Web Vitals, [Largest Contentful Paint](https://web.dev/articles/optimize-lcp) (LCP).

### DIFF
--- a/vale/dictionary.jsonnet
+++ b/vale/dictionary.jsonnet
@@ -44,6 +44,7 @@ local newWord(word, affixes, po) = {
     newWord('Codespaces', '', 'noun') { description: 'GitHub Codespaces https://docs.github.com/en/codespaces/overview', product: true },
     newWord('composable', '', 'adjective'),
     newWord('configure', 'mpDS', 'verb'),
+    newWord('contentful', '', 'adjective'),
     newWord('Couchbase', 'M', 'noun'),
     newWord('Data Firehose', '', 'noun') { Amazon: true, product: true },
     newWord('CPU', 'S', 'noun') { abbreviation: true, established_abbreviation: true },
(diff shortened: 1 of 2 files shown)


vale/dictionary.jsonnet
```diff
@@ -43,6 +43,7 @@ local newWord(word, affixes, po) = {
     newWord('Codespaces', '', 'noun') { description: 'GitHub Codespaces https://docs.github.com/en/codespaces/overview', product: true },
     newWord('composable', '', 'adjective'),
     newWord('configure', 'mpDS', 'verb'),
+    newWord('contentful', '', 'adjective'),
```
clayton-cornell, line 46: Can you add a description field to this one?  This term has a very specific context/use. A quick Google on just `contentful` directs you to the CMS and not to LCP.  Will it be referred to as LCP? Do we need to define an abbreviation as well?

@jdbaldry for something like this, do we have any mechanism in place to ensure that a word like `contentful` stays within the whole phrase it blongs to... like `Largest Contentful Paint`... and isn't used alone?

jdbaldry: Nothing yet but I could implement something that would achieve that I think.

clayton-cornell: I don't know how necessary it would be.  This was a "hmmm I wonder" question. Maybe something to think about if it becomes an issue in the future :-)

heitortsergent: @clayton-cornell would something like this work? 🤓 

```suggestion
    newWord('contentful', '', 'adjective') { description: 'Having content. Used in Web Vital metrics, such as Largest Contentful Paint: https://web.dev/articles/lcp' },
```

clayton-cornell: Yeah that puts it in context.

clayton-cornell (approved):

PR #735, opened 2024-06-12
docs: extend skip rules example

JStickler:
Fleshing this out a little, because for some reason I did NOT understand how to do this when I first read it and just copy/pasted the example.  Which of course, isn't how it works....

### DIFF
--- a/docs/sources/review/lint-prose/_index.md
+++ b/docs/sources/review/lint-prose/_index.md
@@ -26,18 +26,15 @@ Repositories that run Vale in CI include:
 
 ## Skip rules
 
-To skip a rule, enclose the section with HTML comments that first disable, and then re-enable the specific Vale rule. Include the specific rule name in the comment, for example `Grafana.We` or `Grafana.Google.Ellipses`. The `Grafana.Quotes` rule is a made up example rule to illustrate how to disable a rule.
-
-To disable the `Grafana.Quotes` rule:
+To skip a rule, enclose the section with HTML comments that first disable, and then re-enable the specific Vale rule. Include the specific rule name in the comment, for example `Grafana.We` or `Grafana.GoogleEllipses`. The following example shows how to disable the `Grafana.GooglePassive` rule:
 
 ```markdown
-<!-- vale Grafana.Quotes = NO -->
+<!-- vale Grafana.GooglePassive = NO -->
 
-The task title is required.
-The task title succinctly describes the goal to be accomplished as the result of following the instruction.
-The task title contains a verb and an object. For example: "Create a dashboard".
+- [Deprecated content](#deprecation-example): Features that have been deprecated, but still need to be documented for some time.
+- [Configuration options](#configuration-options-example): Features that have several ways they can be configured.
 
-<!-- vale Grafana.Quotes = YES -->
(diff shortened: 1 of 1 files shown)


docs/sources/review/lint-prose/_index.md
```diff
@@ -26,7 +26,7 @@ Repositories that run Vale in CI include:
 
 ## Skip rules
 
-To skip a rule, enclose the section with HTML comments that first disable, and then re-enable the specific Vale rule.
+To skip a rule, enclose the section with HTML comments that first disable, and then re-enable the specific Vale rule. Include the specific rule name in the comment, for example `Grafana.We` or `Grafana.Google.Ellipses`. The `Grafana.Quotes` rule is a made up example rule to illustrate how to disable a rule.
 
 To disable the `Grafana.Quotes` rule:
```
jdbaldry, line 31: Fixing the rule name for [`Grafana/GoogleEllipses`](https://github.com/grafana/writers-toolkit/blob/main/vale/Grafana/GoogleEllipses.yml)

```suggestion
To skip a rule, enclose the section with HTML comments that first disable, and then re-enable the specific Vale rule.
Include the specific rule name in the comment, for example `Grafana.We` or `Grafana.GoogleEllipses`. 
The `Grafana.Quotes` rule is a made up example rule to illustrate how to disable a rule.

To disable the `Grafana.Quotes` rule:
```

jdbaldry:
Thank you for the contribution!

docs/sources/review/lint-prose/_index.md
clayton-cornell: Why use a "fake" rule and have to explain it's fake? Can't we illustrate this with a real rule?

jdbaldry: That makes sense to me. It seems that this was a typo of [`Grafana.GoogleQuotes`](https://github.com/grafana/writers-toolkit/blob/main/vale/google.jsonnet#L31) that has been removed.

jdbaldry: I guess it shouldn't matter really if the rule is real or not after we have the explanation that you need to include the name of the rule you want to disable:

```suggestion
To skip a rule, enclose the section with HTML comments that first disable, and then re-enable the specific Vale rule. Include the specific rule name in the comment, for example `Grafana.We` or `Grafana.Google.Ellipses`.

To disable the `Grafana.Quotes` rule:
```

jdbaldry: I've reopened this conversation because I think it's unresolved.

JStickler: @clayton-cornell If you've got a real example, I'm happy to include it.  I've only skipped a rule once and I didn't think it was a great example to publish.

clayton-cornell: There are several in https://raw.githubusercontent.com/grafana/writers-toolkit/main/docs/sources/write/shortcodes/index.md (and other places).... `<!-- vale Grafana.GooglePassive = NO -->` for example.

chri2547 (approved):
LGTM! Thank you!

jdbaldry (approved):
LGTM, thanks for updating the example.

PR #738, opened 2024-06-13
Add Admin rule that clarifies when it's OK to use admin instead of administrator

jdbaldry:
Outcome from https://docs.google.com/document/d/1uCjtHEy0lpZggkpgsLDjWwxV9K6h7vcsG2dy4TOS6EA/edit#heading=h.valhzaozfw5m


clayton-cornell (approved):
Looks ok to me

PR #745, opened 2024-06-19
Add new words to the dictionary

clayton-cornell:


clayton-cornell:
Should we add the regenerate step to the Add words topic: https://grafana.com/docs/writers-toolkit/review/lint-prose/dictionary/add-words/  ?

jdbaldry (approved):
I think globbing is a hard one to encode in an affix rule because I think the double b comes from the pronunciation and not from the spelling

PR #747, opened 2024-06-20
Update abbreviation rules so that elaborations are in heading exceptions and add web-vital product headings

codecapitano:
I only added those web vitals to the headings file where vale threw an error.

* [Related docs pr](https://github.com/grafana/website/pull/20404)


jdbaldry (commented):
Thank you for the contribution!

This file is generated from the dictionary. I'll retrofit these and regenerate the rule :)

PR #757, opened 2024-06-26
Review 'Markdown guide'

jdbaldry:
- Add semantic line breaks
- Add some more meaningful alt text
- Make some clarifying edits

The rest LGTM

Closes https://github.com/grafana/writers-toolkit/issues/752

### DIFF
--- a/docs/sources/write/markdown-guide/index.md
+++ b/docs/sources/write/markdown-guide/index.md
@@ -10,7 +10,7 @@ keywords:
   - bold
   - tables
   - lists
-review_date: "2024-02-13"
+review_date: "2024-06-26"
 title: Markdown guide
 weight: 300
 ---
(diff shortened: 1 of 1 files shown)


docs/sources/write/markdown-guide/index.md
```diff
@@ -242,8 +242,8 @@ The title text does.
 
 ### Examples
 
-- `![Grafana logo](/link/to/grafanalogo/logo.png "Grafana logo")`
-- `![Example](/static/img/docs/folder_name/alert_test_rule.png "Example title")`
+- `![The Grafana logo](/link/to/grafanalogo/logo.png "Grafana logo")`
+- `![An alert test rule](/static/img/docs/folder_name/alert_test_rule.png "Example title")`
```
clayton-cornell, line 246: What about using the figure shortcode? Which is preferable? markdown or the shortcode?

```
{{< figure alt="<ALT TEXT>"  src="/media/<PATH TO IMAGE>" >}}
```

jdbaldry: Mm, that's an interesting question. I think generally it's not important but you will need to use `figure` if you want captions or additional styling.

jdbaldry: I think that's explained in the next paragraph

clayton-cornell: I clicked expand up.... instead of down. Using the shortcode was/is on my mind because I totally missed that it was available and assumed the markdown was the only way to add images.

clayton-cornell (approved):

PR #779, opened 2024-07-11
Document duplicate heading ids and how to avoid them

jdbaldry:
Duplicate headings only arise when using `docs/shared`.

This PR aims to explain why and how to work around it.

### DIFF
--- a/docs/sources/write/shortcodes/index.md
+++ b/docs/sources/write/shortcodes/index.md
@@ -568,7 +568,7 @@ To trigger a rebuild after changes to a source file, perform a trivial change to
 
 ### Shared files and headings
 
-Hugo renders the shortcode separately to the page. As a result, if the same heading exists in a shared file as well as in the page that includes it, they'll have the same heading identifier. This duplication of heading identifiers breaks the ability link to the headings properly.
+Hugo renders the shortcode separately to the page. As a result, if the same heading exists in a shared file and the page that includes it, they'll have the same heading identifier. This duplication of heading identifiers breaks the ability to link to the headings properly.
 
 To work around this, set a heading identifier in the shared file.
 If there are two shared files with the same heading, you only need to set a heading identifier in one of them.
--- a/docs/sources/write/shortcodes/index.md
+++ b/docs/sources/write/shortcodes/index.md
@@ -568,7 +568,7 @@ To trigger a rebuild after changes to a source file, perform a trivial change to
 
 ### Shared files and headings
 
-Hugo renders the shortcode separately to the page. As a result, if the same heading exists in a shared file as well as in the page that includes it, they'll have the same heading identifier. This duplication of heading identifiers breaks the ability link to the headings properly. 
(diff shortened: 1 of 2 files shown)


docs/sources/write/shortcodes/index.md
```diff
@@ -566,6 +566,14 @@ Hugo doesn't rebuild the destination file when a source file changes on disk.
 To trigger a rebuild after changes to a source file, perform a trivial change to the destination file and save that, too.
 {{< /admonition >}}
 
+{{< admonition type="note" >}}
+Hugo renders the shortcode separately to the page.
+If the same heading exists in a shared file and in the page that includes it, they'll have the same heading identifier.
+
+To work around this, set the heading identifier in the shared file.
+To set a heading identifier, refer to the [Markdown guide](https://grafana.com/docs/writers-toolkit/write/markdown-guide/#identifiers).
+{{< /admonition >}}
```
imatwawana, line 575:
```suggestion
Because Hugo renders the shortcode separately to the page, if the same heading exists in a shared file as well as in the page that includes it, they'll have the same heading identifier. This duplication of heading identifiers breaks the ability link to the headings properly. 

To work around this, set a heading identifier in the shared file.
If there are two shared files with the same heading, you only need to set a heading identifier in one of them.
To set a heading identifier, refer to the [Markdown guide](https://grafana.com/docs/writers-toolkit/write/markdown-guide/#identifiers).
```

imatwawana (commented):
Thank you for tackling this. I've suggested some edits for the shortcodes content. I've also taken it out of an admonition because it's part of properly using a shared file, and that content is a bit long for a note. When I read notes, I think of them as like a side quest where I still need to keep the main campaign in my brain. I think a note this long that includes directions, is a bit too much for a side quest.

docs/sources/write/shortcodes/index.md
clayton-cornell: Could we split line 571 into two sentences?

`Hugo renders the shortcode separately to the page. If the same heading exists in a shared file as well as in the page that includes it, they'll have the same heading identifier.`

imatwawana: I made it once sentence to ensure that those two ideas were tied for the reader. I've made an update to split it up but keep the connection.

docs/sources/write/shortcodes/index.md
```diff
@@ -566,6 +566,14 @@ Hugo doesn't rebuild the destination file when a source file changes on disk.
 To trigger a rebuild after changes to a source file, perform a trivial change to the destination file and save that, too.
 {{< /admonition >}}
 
+### Shared files and headings
+
+Hugo renders the shortcode separately to the page. As a result, if the same heading exists in a shared file as well as in the page that includes it, they'll have the same heading identifier. This duplication of heading identifiers breaks the ability link to the headings properly.
```
chri2547, line 571:
```suggestion
Hugo renders the shortcode separately to the page. As a result, if the same heading exists in a shared file and the page that includes it, they'll have the same heading identifier. This duplication of heading identifiers breaks the ability to link to the headings properly.
```

chri2547 (approved):
one nit

PR #798, opened 2024-07-29
Update Headings.yml

chri2547:
Added new product name "Grafana Cloud Asserts"


jdbaldry:
Thanks for the contribution. I think this is already covered by #796 so let's merge that instead.

PR #800, opened 2024-07-30
Document pyro-flame-graph shortcode

LucasMRC:
Adds documentation entry for the `pyro-flame-graph` shortcode.

Context for the feature is in https://github.com/grafana/website/issues/19745.

### DIFF
--- a/docs/sources/write/shortcodes/index.md
+++ b/docs/sources/write/shortcodes/index.md
@@ -1025,6 +1025,20 @@ You can configure automatic playback with:
 ```markdown
 {{</* youtube id="g97CjKOZqT4" autoplay="true" */>}}
 ```
+## Pyroscope flamegraph
+
+The `pyro-flamegraph` shortcode embeds pyroscope flamegraphs.
+
+| Parameter  | Description                                     | Required |
+| ---------- | ----------------------------------------------- | -------- |
+| `id`       | Flamegraph id                                   | yes      |
+
+The `id` is the value present in the iframe url after the `/shared/` path parameter.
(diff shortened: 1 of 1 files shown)


jdbaldry (commented):
Ah, this PR reminded me that we prefer "flame graph" and not "flamegraph".

Thanks for contributing this. Would you mind if I submitted the website PR to change the shortcode to be `pyro-flame-graph` and then I can also clean up this PR to match that guidance.

LucasMRC:
Sounds perfect!

JStickler:
Curious as to why someone would be inserting an actual graph (vs a screen shot of a flame graph) into the documentation?  I would assume that the data that's generating the graph would need to be publicly visible to all users in all geos?  What is displayed if there is no data? Or the data source is unavailable to a user for some reason (network issues, etc.)?   I'm just not sure what the use case is here....

jdbaldry:
> Curious as to why someone would be inserting an actual graph (vs a screen shot of a flame graph) into the documentation? I would assume that the data that's generating the graph would need to be publicly visible to all users in all geos? What is displayed if there is no data? Or the data source is unavailable to a user for some reason (network issues, etc.)? I'm just not sure what the use case is here....

Good question, I've updated the description to include a link to the issue that prompted the creation of the shortcode.

JStickler:
@jdbaldry thanks for the context.  Should we document that this is intended to be used on Blog posts rather than user docs?

jdbaldry:
> @jdbaldry thanks for the context. Should we document that this is intended to be used on Blog posts rather than user docs?

@jstickler Although the shortcode was originally implemented to allow the migration of Pyroscope blog posts, are there reasons we should discourage use of this in documentation?

@LucasMRC do you know of any reasons to not use this in documentation?

LucasMRC:
> @LucasMRC do you know of any reasons to not use this in documentation?

Nope, I don't know of any specific reason why not to use it in docs.

PR #805, opened 2024-08-07
Add grammar rules for drop-down and dataset spelling

imatwawana:


jdbaldry (approved):
Thanks, I'm gonna update the Vale rule to more correctly report this too.

PR #819, opened 2024-08-28
Document how to structure references

jdbaldry:


docs/sources/write/style-guide/write-for-developers/index.md
```diff
@@ -85,6 +85,14 @@ Additional tips:
   For example, don't change `MyEvent` to `MyEvents`; refer to the `MyEvent` objects.
 - If the element does some sort of action, start the first sentence of the description with an action verb.
 
+### Reference structure
+
+Organize references alphabetically, organizing by frequency of use is too subjective.
+
+If the reference has required and optional components, document them separately.
```
clayton-cornell, line 92: Do we interpret this as... put the required at the top of the list, or as a separate section?

I'm not yet 100% on putting required as a separate section.

As a separate section it could look like this:
image

jdbaldry: Yeah, I'm not certain either. Might be easiest to resolve this discussion in Slack and then update this PR with the outcome and reasons.

chri2547: Maybe add required or optional in the description, as the first bit of information. Like "Required. Handler from an..."? I think that's how I've done this in the past.

chri2547 (approved):

jdbaldry:
Gonna get this in to stop it hanging around but I'll happily review any iterations on the advice

PR #822, opened 2024-08-30
Fix link to guidance for dark and light mode

jdbaldry:
The original link goes to a different section that doesn't contain the guidance referenced.


clayton-cornell (approved):
LGTM

PR #823, opened 2024-08-30
Remove all non-semantic uses of <br>

jdbaldry:
`<br>` elements shouldn't be used to present information in most cases. Certainly not to break paragraphs in tables.
Semantic HTML is more accessible, improving readability in screen readers and other tools.
I want to improve this practice across Grafana and we need to lead by example.

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element 

https://developers.google.com/style/tables#multi-paragraph-table-cells


clayton-cornell (approved):
LGTM

PR #826, opened 2024-09-03
Link to Google documentation style guide for table style

jdbaldry:
Also specifically note not to use `<br>` HTML elements.


clayton-cornell (approved):
Makes sense. LGTM

Issue #600, opened 2024-04-02
Demote some Vale warnings to suggestions


jdbaldry:
Anything with regular false positives.

- **Grafana.GoogleOxfordComma**
  - https://github.com/grafana/website/pull/19030/files#r1538318854
  - TODO compare the implementation with https://github.com/redhat-documentation/vale-at-red-hat/blob/main/.vale/styles/RedHat/OxfordComma.yml

PR #827, opened 2024-09-03
Demote Grafana.GoogleOxfordComma to suggestion

jdbaldry:
There are too many false positives and this rule is easier to spot by eye than with regular expressions.

Closes https://github.com/grafana/writers-toolkit/issues/600


clayton-cornell (approved):
LGTM

PR #842, opened 2024-09-13
Improve rendering of substitution and existence rules

jdbaldry:
Render the swaps or tokens that trigger rules.

Screenshot 2024-09-13 at 10 41 04

Screenshot 2024-09-13 at 10 41 20


docs/sources/review/lint-prose/rules.md
```diff
@@ -122,77 +139,151 @@ Extends: existence
...
+| `mail(?:m[ae]n\|wom[ae]n)`     | `mail carriers`               |
+| `man and wife`                 | `husband and wife`            |
+| `man enough`                   | `strong enough`               |
+| `mankind`                      | `human kind`                  |
+| `manmade`                      | `manufactured`                |
+| `manpower`                     | `personnel`                   |
+| `men and girls`                | `men and women`               |
+| `middle(?:m[ae]n\|wom[ae]n)`   | `intermediary`                |
+| `news(?:m[ae]n\|wom[ae]n)`     | `journalist(s)`               |
+| `ombuds(?:man\|woman)`         | `ombuds`                      |
```
clayton-cornell, line 196: Jut wondering... is maybe `ombudsperson` more usable here? I see it used more often in Canada as ombudsperson vs just ombuds. I doubt it'll land in the tech docs anyway... so not worth a lot of bikeshedding. :-) Just curious if it's regional.

clayton-cornell (approved):
Nothing jumped out as a mistake.. LGTM.. One minor question.

docs/sources/review/lint-prose/rules.md
jdbaldry: Oh yeah I think so too. These are some inherited rules from the Google style guide Vale implementation but we can update it.

PR #847, opened 2024-09-17
Fix some visual bugs in rule rendering

jdbaldry:
- Ensure all substitutions have exactly two format codes.
- Escape pipes in replacement regular expressions.
- Only transform placeholder variables and not all text between `<` and `>`


clayton-cornell (approved):
LGTM

PR #853, opened 2024-09-26
Update vale/dictionary with PLZ abbreviation

yorugac:
I have been getting a vale spelling error in PR: https://github.com/grafana/website/pull/21811

We often use this abbreviation "PLZ" both internally and externally, and it is unlikely to be misread, at least within Grafana docs context.


clayton-cornell (approved):
Looks OK to me.  @jdbaldry I think everything is formatted correctly?

We need to make sure the term is defined within the body text of a topic. German speaking readers will likely first associate PLZ with Postleitzahl (German abbreviation for for Post Code).

jdbaldry:
Yup, this looks good to me, and given that it's not marked as an "established abbreviation", there will be a Vale rule to ensure that we define PLZ in the first use.

PR #858, opened 2024-10-01
Add TypeScript script that replaces website-sync

jdbaldry:


jdbaldry:
Not needed anymore, we pull not push for the sync

PR #862, opened 2024-10-01
Docs: Add kebab case and single pane of glass to word list

imatwawana:


docs/sources/write/style-guide/word-list/index.md
```diff
@@ -132,7 +132,17 @@ Use this rather than _hold the pointer over_ or _point to_.
 <!--
 ## I
 ## J
+-->
+
 ## K
+
+### kebab case
+
+Use this to refer to the casing in which lower case words are connected by dashes or hyphens.
```
clayton-cornell, line 141: Is this really "casing"?  This is more to do with replacing spaces with dashes to "form a multi-word identifier".

Maybe a rephrase to something more like this?
```
Use this to refer to the naming convention where spaces between words are replaced with dashes to form a multi-word identifier.
```

imatwawana, line 141:
```suggestion
Use this to refer to the naming convention where spaces between lower case words are replaced with dashes.```

clayton-cornell (approved):
LGTM

Issue #883, opened 2024-10-22
Review https://github.com/grafana/writers-toolkit/blob/main/docs/sources/contribute/_index.md


jdbaldry:
Behavior has changed in https://github.com/grafana/website/pull/22302/

PR #884, opened 2024-10-22
Update documentation in preparation for changes to feedback links

jdbaldry:
Changes incoming in https://github.com/grafana/website/pull/22302/

Closes https://github.com/grafana/writers-toolkit/issues/883

I may have had a bit too much fun with shortcodes. Perhaps I should just update the wording instead?

### DIFF
--- a/docs/sources/contribute/_index.md
+++ b/docs/sources/contribute/_index.md
@@ -37,7 +37,7 @@ Or you can email [`docs@grafana.com`](mailto:docs@grafana.com) directly:
 
 The Grafana Labs documentation team checks the email inbox regularly and responds to emails in a timely fashion.
 
-## {{% translate.inline "docs_feedback_contribute" /%}}
+## {{% translate.inline "docs_feedback_contribute" / %}}
 
 If you want to recommend a small change, such as suggesting a correction to a topic, you can edit the topic directly in GitHub.
 
--- a/docs/sources/contribute/_index.md
+++ b/docs/sources/contribute/_index.md
@@ -21,14 +21,14 @@ weight: 200
 There are a number of different ways to contribute to documentation at Grafana Labs.
 Choose the one that applies best and learn how you can engage with the documentation team and help by contributing your ideas.
 
-## {{% translate.inline "docs_feedback_report" %}}{{ markdownify (T (.Get 0)) }}{{% /translate.inline %}}
+## {{% translate "docs_feedback_report" %}}
 
 Report a problem when you want to make a suggestion or provide feedback about a topic, but don't want to make the change yourself.
 
 Reporting a problem gives you the freedom to express your ideas without committing language.
 Your suggestion can reflect a small change to wording or can reflect larger, more substantive changes.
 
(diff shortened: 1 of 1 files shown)


docs/sources/contribute/_index.md
```diff
@@ -21,26 +21,23 @@ weight: 200
 There are a number of different ways to contribute to documentation at Grafana Labs.
 Choose the one that applies best and learn how you can engage with the documentation team and help by contributing your ideas.
 
-## Report a problem
+## {{% translate.inline "docs_feedback_report" %}}{{ markdownify (T (.Get 0)) }}{{% /translate.inline %}}
```
robbymilo, line 24: Any reason we cannot just use `{{% translate.inline "docs_feedback_report" /%}}` here and below? I am not sure we want to promote the use of `markdownify` in content. 

Also, TIL about the translate shortcode, should we add to /docs/writers-toolkit/contribute/?

jdbaldry: Good point about the use of `markdownify` and I'm glad you asked about it because I'm sure that we don't want to be encouraging it now that I've thought about it.

The reason was I spotted the Markdown in one of the translations https://github.com/grafana/website/pull/22302/files#diff-360ec96d474fa3fc5b356989607caae93e55b6183677d569dc68ea547427d8cdR57 which is then rendered in https://github.com/grafana/website/pull/22302/files#diff-b10673a8041efa210d39d2de3910e39470d74915ee573ed2fd1da4ca5a06afccR23.

I think we probably want to keep the translation strings free of Markdown too?

jdbaldry: I've also actually cheated here by using an [inline shortcode](https://gohugo.io/templates/shortcode/#inline-shortcodes).

If we like the idea of it, I should really make it a proper shortcode and document it as you've proposed

robbymilo: Ah yes, fixed in https://github.com/grafana/website/pull/22302/commits/842226cebc8f9e0c3648c759ce09da35900aff8b.

docs/sources/contribute/_index.md
```diff
@@ -58,7 +55,7 @@ Small changes might include:
 
 ### Find the source repository
 
-Technical documentation published from public projects have a **Suggest an edit** link with a pencil icon.
+Technical documentation published from public projects have a **{{< translate "docs_feedback_contribute" >}}** link with a pencil icon.
```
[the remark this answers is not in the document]
jdbaldry: Yeah that's a good idea. I'm not sure how to express the virtual latest idea in a way that would make sense to readers. Perhaps we can ignore it and just talk about "latest"?

```suggestion
The "latest" and "next" versions of documentation published from public projects have a **{{< translate "docs_feedback_contribute" >}}** link with a pencil icon.
```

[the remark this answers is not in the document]
robbymilo: Yep that works.

robbymilo (approved):

Issue #870, opened 2024-10-10
Clarify what the presence of front matter page labels means


jdbaldry:
https://grafana.com/docs/writers-toolkit/write/front-matter/#labels

A page should have all labels that are applicable. Presence of a label indicates that the page has at least some content relevant for that label.

Absence of a label means that the page has no content relevant for that label.

PR #885, opened 2024-10-24
Document how to use labels

jdbaldry:
Closes https://github.com/grafana/writers-toolkit/issues/870

### DIFF
--- a/docs/sources/write/front-matter/index.md
+++ b/docs/sources/write/front-matter/index.md
@@ -306,7 +306,7 @@ For `labels.products`, the supported values and the resulting published labels a
 - `oss`: <span class="badge docs-labels__product-oss docs-labels__item">Open source</span>
 
 You should use all labels that apply to the page's content.
-If a page has some open source content, and some Grafana Cloud content, set both labels.
+If a page has some open source content and some Grafana Cloud content, set both labels.
 
 For `labels.stages`, the supported values and the resulting published labels are as follows:
 
--- a/docs/sources/write/front-matter/index.md
+++ b/docs/sources/write/front-matter/index.md
@@ -297,7 +297,7 @@ Ideally, use single terms as opposed to phrases.
 ### Labels
 
 Use the `labels` key to add one or more values that you want to appear before the topic title on the published page.
-The website only supports certain labels values.
+The website only supports certain label values.
 
 For `labels.products`, the supported values and the resulting published labels are as follows:
 
--- a/docs/sources/write/front-matter/index.md
+++ b/docs/sources/write/front-matter/index.md
@@ -297,7 +297,7 @@ Ideally, use single terms as opposed to phrases.
 ### Labels
 
 Use the `labels` key to add one or more values that you want to appear before the topic title on the published page.
-Only certain labels are supported.
+The website only supports certain labels values.
 
 For `labels.products`, the supported values and the resulting published labels are as follows:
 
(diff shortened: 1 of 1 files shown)


docs/sources/write/front-matter/index.md
```diff
@@ -297,20 +297,26 @@ Ideally, use single terms as opposed to phrases.
 ### Labels
 
 Use the `labels` key to add one or more values that you want to appear before the topic title on the published page.
-Only certain labels are supported.
+The website only supports certain labels values.
```
robbymilo, line 300:
```suggestion
The website only supports certain label values.
```

robbymilo (approved):
One nit (I think), otherwise looks good.

docs/sources/write/front-matter/index.md
```diff
@@ -297,20 +297,26 @@ Ideally, use single terms as opposed to phrases.
...
+The website only supports certain label values.
 
 For `labels.products`, the supported values and the resulting published labels are as follows:
 
 - `cloud`: <span class="badge docs-labels__product-cloud docs-labels__item">Grafana Cloud</span>
 - `enterprise`: <span class="badge docs-labels__product-enterprise docs-labels__item">Enterprise</span>
 - `oss`: <span class="badge docs-labels__product-oss docs-labels__item">Open source</span>
 
+You should use all labels that apply to the page's content.
+If a page has some open source content, and some Grafana Cloud content, set both labels.
```
jdbaldry, line 309:
```suggestion
If a page has some open source content and some Grafana Cloud content, set both labels.
```

docs/sources/write/front-matter/index.md
```diff
@@ -297,20 +297,26 @@ Ideally, use single terms as opposed to phrases.
 ### Labels
 
 Use the `labels` key to add one or more values that you want to appear before the topic title on the published page.
-Only certain labels are supported.
+The website only supports certain label values.
```
clayton-cornell, line 300: Mmmm is this true? The Alloy docs uses label values that are not in this list. You can set the label value (text?) to anything you want.  There are a limited set of label colours available to use...

jdbaldry: Do you have a source link handy?

clayton-cornell: The badge colours are defined here (at least this is the reference I used): https://github.com/grafana/website/blob/b972ab4784f5ec5ccd48b9eb0a6b1a7cb1f5ac9b/assets/scss/legacy/modules/_docs.scss#L1681

So the defined badge colours are

* stage = violet-4
* support = gray-1
* product-oss = pink-4
* product-cloud = orange-4
* product-cloud-all = red-4
* product-cloud-pro = orange-4
* product-cloud-advanced = blue-4
* product-enterprise = purple-4

You can use these in any combination in the class. ie the color for `product-cloud` does not need to only be set against `Grafana Cloud` label text... 

For example in Alloy docs, you will see: `<span class="badge docs-labels__stage docs-labels__item">Public preview</span>`   Using the stage label colour to label a topic as `Public Preview`.

jdbaldry: Ah yeah, but you're kinda cheating cause you're using HTML in the Markdown (which I may have encouraged you to do) to simulate badges, you're not actually using the labels front matter if I understand correctly

clayton-cornell: Mmmm true... this isn't being used in the metadata block.  I'd 100% prefer to use it correctly.  Something for the future improvement list?

PR #891, opened 2024-10-29
Use multi-platform Vale image

jdbaldry:


jdbaldry:
GitHub Actions outage is causing the checks to get stuck. 

They haven't announced it yet but all my jobs are in a queue for many minutes

PR #904, opened 2024-11-26
add deploy preview workflow

robbymilo:
for https://github.com/grafana/website/issues/10101

This PR deploys a publicly available preview of the output of the `make docs` command for pull requests. The preview is then deleted when the PR is merged or closed.

### To test
1. Create a branch from this PR
1. Open a PR
1. Test the GitHub comment link
1. Test the link in the GitHub commit checks area
1. Push a new test commit
1. Once the deploy preview is updated, make sure the commit is reflected: 
```
curl -Is https://deploy-preview-writers-toolkit-{replace-with-pr}-zb444pucvq-vp.a.run.app | grep build
```
The output should be the same commit sha as the commit you pushed.
1. Again test the GitHub link (you should see the comment has been edited)
1. Close the PR
1. Confirm the comment is updated with "Deploy preview deleted"
2. Confirm the preview is actually deleted - it should 404.

Before merge todo:
- [x] change clone ref branch
- [x] change workflow ref branch

### DIFF
--- a/.github/workflows/deploy-pr-preview.yml
+++ b/.github/workflows/deploy-pr-preview.yml
@@ -1,0 +1,20 @@
+name: Deploy pr preview
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - closed
+
+jobs:
+  deploy-pr-preview:
+    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@robbymilo/deploy-preview
+    with:
+      sha: ${{ github.event.pull_request.head.sha }}
+      branch: ${{ github.ref_name }}
+      event_number: ${{ github.event.number }}
+      title: ${{ github.event.pull_request.title }}
+      repo: writers-toolkit
+      website_directory: content/docs/writers-toolkit
+      relative_prefix: /docs/writers-toolkit/
--- a/.github/workflows/deploy-pr-preview.yml
+++ b/.github/workflows/deploy-pr-preview.yml
@@ -6,6 +6,8 @@ on:
       - opened
       - synchronize
       - closed
+  paths:
+    - "docs/sources/**"
 
 jobs:
   deploy-pr-preview:
--- a/.github/workflows/deploy-pr-preview.yml
+++ b/.github/workflows/deploy-pr-preview.yml
@@ -14,7 +14,7 @@ jobs:
     uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@robbymilo/deploy-preview
     with:
       sha: ${{ github.event.pull_request.head.sha }}
-      branch: ${{ github.ref_name }}
+      branch: ${{ github.head_ref }}
       event_number: ${{ github.event.number }}
       title: ${{ github.event.pull_request.title }}
       repo: writers-toolkit
--- a/.github/workflows/deploy-pr-preview.yml
+++ b/.github/workflows/deploy-pr-preview.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   deploy-pr-preview:
-    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@robbymilo/deploy-preview
+    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@main
     with:
       sha: ${{ github.event.pull_request.head.sha }}
       branch: ${{ github.head_ref }}
--- a/.github/workflows/deploy-pr-preview.yml
+++ b/.github/workflows/deploy-pr-preview.yml
@@ -6,8 +6,8 @@ on:
       - opened
       - synchronize
       - closed
-  paths:
-    - "docs/sources/**"
+    paths:
+      - "docs/sources/**"
 
 jobs:
   deploy-pr-preview:
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -1,0 +1,194 @@
+name: deploy-preview
+
+on:
+  workflow_call:
+    inputs:
+      sha:
+        required: true
+        type: string
+      branch:
+        required: true
+        type: string
+      event_number:
+        required: true
+        type: string
+      title:
+        required: true
+        type: string
+      repo:
+        required: true
+        type: string
+      source_directory:
+        default: docs/sources
+        type: string
+      website_directory:
+        required: true
+        type: string
+      relative_prefix:
+        required: true
+        type: string
+
+env:
+  CLOUD_RUN_REGION: us-south1
+
+permissions:
+  contents: read
+  id-token: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ inputs.repo }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  deploy-preview:
+    permissions: write-all
+    runs-on: ubuntu-latest
+    steps:
+      - name: Find comment
+        uses: peter-evans/find-comment@3eae4d37986fb5a8592848f6a574fdf654e61f9e # v3.1.0
+        id: fc
+        with:
+          issue-number: ${{ inputs.event_number }}
+          comment-author: "github-actions[bot]"
+          body-includes: Deploy preview available
+
+      - name: Update comment with in-progress deploy preview
+        if: steps.fc.outputs.comment-id != '' && github.event.action != 'closed'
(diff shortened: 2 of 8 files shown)


.github/workflows/deploy-pr-preview.yml
```diff
@@ -0,0 +1,22 @@
...
+      - closed
+    paths:
+      - "docs/sources/**"
+
+jobs:
+  deploy-pr-preview:
+    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@robbymilo/deploy-preview
+    with:
+      sha: ${{ github.event.pull_request.head.sha }}
+      branch: ${{ github.ref_name }}
```
jdbaldry, line 17: Same change as https://github.com/grafana/website/commit/17a0f7f6c3529dd9bc13911fb8332981e5aba128 (whitespace mistake notwithstanding)

```suggestion
      branch: ${{ github.head_ref }}
```

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,194 @@
...
+  contents: read
+  id-token: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ inputs.repo }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  deploy-preview:
+    permissions: write-all
```
jdbaldry, line 44: I know it may be a pain to whittle down the list but do we really need all the write permissions?

From my reading we need `id-token: write` for the auth parts, and `statuses: write` for the commit status updates, and `pull-requests: write` to put the comments.

If we do list the permissions explicitly, it's probably worth commenting what purpose each permission serves perhaps with a link or mention of the step that needs it so we can remember in the future.

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,194 @@
...
+      # get the Dockerfile and nginx conf
+      - name: Sparse checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        if: github.event.action != 'closed'
+        with:
+          repository: "grafana/writers-toolkit"
+          ref: "robbymilo/deploy-preview"
+          path: deploy-preview-files
+          sparse-checkout: |
+            deploy-preview
```
jdbaldry, line 78: I guess we can change this after the two development branches are merged?

jdbaldry (commented):
This is an absolutely incredible feature. Thank you so much for implementing it.

I know you've worked with the Platform team already but I'd like to see if I get one of them to take a look through to check for Actions best practices, primarily from a security perspective, so that we can confidently propagate this across other repositories.

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,194 @@
...
+        uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043 # v4.0.0
+        with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
+          issue-number: ${{ inputs.event_number }}
+          edit-mode: append
+          append-separator: newline
+          body: |
+            :building_construction: Updating deploy preview...
+
+      - uses: actions/checkout@v4
```
zerok, line 66: Please pin this version :)

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,194 @@
...
+        if: github.event.action != 'closed'
+        with:
+          repository: "grafana/writers-toolkit"
+          ref: "robbymilo/deploy-preview"
+          path: deploy-preview-files
+          sparse-checkout: |
+            deploy-preview
+
+      - name: Build website
+        if: github.event.action != 'closed'
```
zerok, line 81: Would synchronize/opened be enough here?

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,194 @@
...
+permissions:
+  contents: read
+  id-token: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ inputs.repo }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  deploy-preview:
```
zerok, line 43: Would it perhaps make the whole workflow more readable if "deploy-preview" and "delete-preview" were separate jobs?

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,194 @@
...
+        with:
+          project_id: grafanalabs-dev
+
+      - name: Delete deploy preview
+        if: github.event.action == 'closed'
+        run: |
+          gcloud run services delete deploy-preview-${{ inputs.repo }}-${{ inputs.event_number }} --region=${{ env.CLOUD_RUN_REGION }} --project=grafanalabs-dev --quiet
+
+      - name: Send commit status
+        uses: ouzi-dev/commit-status-updater@v2
```
zerok, line 160: Pin the version please :) (also on other calls above)

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,194 @@
...
+
+      - name: Build website
+        if: github.event.action != 'closed'
+        shell: bash
+        run: |
+          mkdir -p ${PWD}/dist
+          docker run -v ${PWD}/dist:/hugo/dist -v "${PWD}/${{ inputs.source_directory }}:/hugo/${{ inputs.website_directory }}" --rm grafana/docs-base:latest /bin/bash -c 'HUGO_SSI=false hugo --environment=docs --destination=dist/ --baseURL= --minify'
+          printf "%s" "add_header 'Build' '"${{ inputs.sha }}"';" > build.conf
+
+      - uses: google-github-actions/auth@6fc4af4b145ae7821d527454aa9bd537d1f2dc5f # v2.1.7
```
zerok, line 88: Couldn't you use https://github.com/grafana/shared-workflows/tree/main/actions/push-to-gar-docker for this and the following steps without requiring a custom service account?

robbymilo: push-to-gar uses login-to-gar, which then uses the service account from the `construct-service-account` step 

The service account in login-to-gar (https://github.com/grafana/shared-workflows/blob/main/actions/login-to-gar/action.yaml#L27), which is called from `push-to-gar`, is hard coded from the output of the `construct-service-account` step.

.github/workflows/deploy-preview.yml
robbymilo: Yep, added as a todo.

.github/workflows/deploy-preview.yml
robbymilo: I toyed with that idea but the workflow became very verbose with quite a few steps duplicated.

.github/workflows/deploy-preview.yml
robbymilo: I actually added the files to this branch so we only one PR is needed.

.github/workflows/deploy-preview.yml
jdbaldry: Thanks for exploring that :)

.github/workflows/deploy-preview.yml
jdbaldry: Happy to explore this in a follow up. I don't mind that the same service account handles all the aspects of this service.

jdbaldry (approved):
Looks so good man!

.github/workflows/deploy-preview.yml
zerok: The service account in the shared-workflow is currently just an implementation detail and IIRC @dsotirakis is investigating if we can completely get rid of it 🙂 That was the main reason for me asking about this here but it's absolutely not required 🙂

zerok (approved):
Thank you for porting this over ❤️

.github/workflows/deploy-preview.yml
jdbaldry: > The service account in the shared-workflow is currently just an implementation detail and IIRC @dsotirakis is investigating if we can completely get rid of it 🙂 That was the main reason for me asking about this here but it's absolutely not required 🙂

Good to know! I'm subscribed to all activity on the grafana/shared-workflows repository so if something moves there, we'll know :) 

Thanks for clarifying

jdbaldry (approved):

.github/workflows/deploy-preview.yml
robbymilo: Happy to come back and update if we can get rid of it!

PR #908, opened 2024-11-28
only checkout necessary directories

robbymilo:


.github/workflows/deploy-preview.yml
```diff
@@ -66,6 +66,9 @@ jobs:
 
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         if: github.event.action == 'opened' || github.event.action == 'synchronize'
+        with:
+          sparse-checkout-cone-mode: false # exclude root files
+          sparse-checkout: docs
```
jdbaldry, line 71: We can make this configurable in the future if we like but every standard project and every one that I can think of (ignoring plugins) uses this structure.

.github/workflows/deploy-preview.yml
```diff
@@ -78,6 +81,16 @@ jobs:
...
+      # sparse checkout with cone mode disabled includes root files, even when using exclusions
+      # see https://github.com/actions/checkout/issues/1430#issuecomment-1756326892
+      - name: Keep only necessary files
+        if: github.event.action == 'opened' || github.event.action == 'synchronize'
+        shell: bash
+        run: |
+          find . -mindepth 1 -not \( -path "./${{ inputs.source_directory }}" -o -path "./${{ inputs.source_directory }}/*" -o -path "./deploy-preview-files" -o -path "./deploy-preview-files/*" \) -exec rm -rf {} +
+          ls -al
+          ls -al $trimmed_value
+
```
jdbaldry, line 93: Can we also just consider trashing all files that aren't our docs directory? 

Or perhaps another way is to do all the checking out into the default workspace but then copying what we want into a new subdirectory.

The workspace tree may then look like:

```
.
├── build
│   ├── deploy-preview-files
│   ├── dist
│   └── docs
│       └── sources
├── deploy-preview-files
└── grafana
```

jdbaldry: In that previous comment, the `grafana` directory contains the grafana/grafana repo checkout (sparse or otherwise). The `build` directory is where we copy the `deploy-preview-files;` from the grafana/writers-toolkit checkout and `docs/sources/` from the grafana/grafana checkout.

We then switch to that `build` directory before running all the rest of the action

.github/workflows/deploy-preview.yml
robbymilo: Great, that means I can get rid of trimming the `source_directory`.

.github/workflows/deploy-preview.yml
robbymilo: Yep, that is what I'm trying with the find command. There is probably a cleaner way.

jdbaldry: Oh right yeah, the `find` finds all files that aren't in our preferred list and removes them. That may be easiest actually because I recall that performing multiple steps in a subdirectory of an action is really verbose.

I think the find command along with maybe capturing the "to keep" directories in a Bash array for readability would be a good way to go.

I'm happy to pick this up tomorrow if you like. I came across this PR because I was hoping to give you a contribution for your review tomorrow but wanted to make sure that I wasn't stepping on your toes haha.

robbymilo: Take a look at https://github.com/grafana/writers-toolkit/pull/908/commits/09544788c1d02141d37424170561a0d0c79b96e1 and its run in the grafana repo: https://github.com/grafana/grafana/actions/runs/12081483043/job/33690724463

PR #927, opened 2024-12-13
Error on <br> tags in source files

jdbaldry:
They are almost never semantically correct: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element

### DIFF
--- a/vale/Grafana/Paragraphs.yml
+++ b/vale/Grafana/Paragraphs.yml
@@ -1,0 +1,7 @@
+extends: script
+scope: raw
+level: error
+link: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element
+message: |
+  br elements must be used only for line breaks that are actually part of the content, as in poems or addresses.
+script: Paragraphs.tengo
--- a/vale/config/scripts/Paragraphs.tengo
+++ b/vale/config/scripts/Paragraphs.tengo
@@ -1,0 +1,36 @@
+DEBUG := false
+
+fmt := import("fmt")
+text := import("text")
+
+matches := []
+ruleActive := true
+
+cursor := 0
+for line in text.split(scope, "\n") {
+  if DEBUG {
+    fmt.println(text.pad_left(text.itoa(cursor), 3, " "), ": ",  text.quote(line), " ", len(line))
+  }
+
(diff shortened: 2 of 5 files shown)


clayton-cornell:
Tested and it works as expected.

But.... should this maybe be implemented as a **Warning** to start with? I have a feeling that it will block a LOT of PRs if it's set to error (based on a quick grep of a few of the doc repositories).  Or should we fix the repos first and then merge this PR?

jdbaldry:
> But.... should this maybe be implemented as a Warning to start with? I have a feeling that it will block a LOT of PRs if it's set to error (based on a quick grep of a few of the doc repositories). Or should we fix the repos first and then merge this PR?

That's an important consideration but we don't actually run Vale in CI anywhere except in this repository where both warnings and errors block the builds.

clayton-cornell (approved):

vale/Grafana/Paragraphs.yml
```diff
@@ -0,0 +1,7 @@
+extends: script
+scope: raw
+level: error
+link: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element
+message: |
+  br elements must be used only for line breaks that are actually part of the content, as in poems or addresses.
```
jdbaldry, line 6:
```suggestion
  `br` elements must be used only for line breaks that are actually part of the content, as in poems or addresses.
  
  Outside of tables, replace the use of `br` elements with semantic Markdown.
  In tables, use semantic HTML elements like `p` for paragraphs, and `ul` or `ol` for lists.
```

jdbaldry (commented):
Proposing a better message

vale/Grafana/Paragraphs.yml
clayton-cornell: > In tables, use semantic HTML elements like `p` for paragraphs, and `ul` or `ol` for lists.

Should this maybe include the `<` and `>`?

jdbaldry: I'm open to it, I probably switch every time I write something.

Google seems to omit the brackets: https://developers.google.com/style/semantic-tagging
MDN includes them: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p

robbymilo (approved):

PR #970, opened 2025-01-15
Docsc: Add CSV file and files to word list

imatwawana:
This PR
- Adds _CSV file_ and _CSV files_ as suggested replacements for _CSV_ and _CSVs_ in Word List
- Removes _CSVs_ from dictionary

Should have done this instead of adding _CSVs_ to the dictionary per [PR 967](https://github.com/grafana/writers-toolkit/pull/967) 🙃

### DIFF
--- a/vale/Google/WordList.yml
+++ b/vale/Google/WordList.yml
@@ -43,7 +43,6 @@ swap:
   Cloud: Google Cloud Platform|GCP
   Container Engine: Kubernetes Engine
   content type: media type
-
   curated roles: predefined roles
   data are: data is
   Developers Console: Google API Console|API Console
--- a/vale/Google/WordList.yml
+++ b/vale/Google/WordList.yml
@@ -43,6 +43,7 @@ swap:
   Cloud: Google Cloud Platform|GCP
   Container Engine: Kubernetes Engine
   content type: media type
+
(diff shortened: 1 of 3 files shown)


clayton-cornell (approved):
My first thoguht was no wait, that's not what's documented here: https://grafana.com/docs/writers-toolkit/review/lint-prose/dictionary/add-words/ but...

You explicitly want to add `CSV file` and `CSV files` to the wordlist... so this makes sense.

imatwawana:
> My first thought was no wait, that's not what's documented here: [grafana.com/docs/writers-toolkit/review/lint-prose/dictionary/add-words](https://grafana.com/docs/writers-toolkit/review/lint-prose/dictionary/add-words/) but...
> 
> You explicitly want to add `CSV file` and `CSV files` to the wordlist... so this makes sense.

Yeah, I want to be able to have the suggestions come up for folks, because that would have helped me in this instance.

vale/Google/WordList.yml
```diff
@@ -43,6 +43,7 @@ swap:
   Cloud: Google Cloud Platform|GCP
   Container Engine: Kubernetes Engine
   content type: media type
+
```
imatwawana, line 46:
```suggestion
```

PR #973, opened 2025-01-15
Revert "Docsc: Add CSV file and files to word list"

jdbaldry:
Reverts grafana/writers-toolkit#970

### DIFF
--- a/vale/Grafana/styles/Grafana/WordList.yml
+++ b/vale/Grafana/styles/Grafana/WordList.yml
@@ -40,8 +40,6 @@
   "blacklists": "blocklists"
   "check[- ]box": "checkbox"
   "content type": "media type"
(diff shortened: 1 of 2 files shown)


clayton-cornell (commented):
So what's the right way to do this?  If CSV should be replaced with `CSV files` isn't adding `S` to the word definition only going to make `CSVs` a valid word?

jdbaldry:
Yeah, the right way is to make sure the regular expression only matches CSV when it's not followed by " files".

We may want to consider also whether we need a dictionary entry for CSVs anyway since it's a valid pluralization even if we want to encourage CSV files when people mean that.

jdbaldry:
I can open a PR to add CSV -> CSV file and show how that regular expression works.

PR #987, opened 2025-01-21
Add netlink Vale word

grafsean:
Used in https://github.com/grafana/beyla/pull/1545

- [ ] I've used a relevant pull request (PR) title.
- [ ] I've added a link to any relevant issues in the PR description.
- [ ] I've checked my changes on the deploy preview and they look good.
- [ ] I've added an entry to the [What's new](https://github.com/grafana/writers-toolkit/blob/main/docs/sources/whats-new.md) page (only required for notable changes).

PR #991, opened 2025-01-22
Add Figma to dictionary

jdbaldry:


docs/sources/review/lint-prose/rules.md
```diff
@@ -801,6 +802,7 @@ _`<CURRENT TEXT>`_ was matched by one or more of the following regular expressio
 - `Mimir's`
 - `Moodle's`
 - `MySQL's`
+- `Netlink's`
```
clayton-cornell, line 805: You added `Netlink` here and in a couple other places, but I don't see it in the n.jsonnet file or in other places like Wordlist.yml. Is this intentional? Or was it missed in a couple palces?

jdbaldry: Missed in https://github.com/grafana/writers-toolkit/pull/987 because I had a bug in the build script so it misunderstood which dependencies it need to rebuild.

I'm being lazy here including the fixed files in this PR rather than having opened a separate one. Happy to be held to account and do things properly.

clayton-cornell (approved):

docs/sources/review/lint-prose/rules.md
clayton-cornell: Naaah... I was just trying to do a proper review and check things. All good.

PR #1003, opened 2025-01-24
Add Kusto and KQL definitions

aangelisc:
Adding definitions for Kusto and KQL


jdbaldry (approved):
Thanks for adding this!

After merge, you'll no longer see this error on your PR but you may have to resolve the conversation manually to hide the comment from the bot.

PR #1014, opened 2025-01-30
Replace individual inputs with JSON string array of inputs

jdbaldry:
This lets us build multiples sources in a single deploy preview as demonstrated in https://github.com/grafana/mimir/pull/10580.

I also think it allows us to mount source projects into Grafana Cloud documentation directly instead of a publishing directory although it doesn't use Hugo mounts for that.

Let me know if you'd like me to test anything specifically out for you so you can be confident in the changes.

This can be rolled out safely as tested in:
- [Legacy inputs](https://github.com/grafana/fleet-management/actions/runs/13409179066/workflow?pr=490)
- [New inputs](https://github.com/grafana/writers-toolkit/actions/runs/13409334007/job/37455648763?pr=1014)

- [x] I've added a link to any relevant issues in the PR description.
- [x] I've checked my changes on the deploy preview and they look good.

### DIFF
--- a/.github/workflows/deploy-pr-preview.yml
+++ b/.github/workflows/deploy-pr-preview.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   deploy-pr-preview:
     if: "!github.event.pull_request.head.repo.fork"
-    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@jdb/deploy-preview
+    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@main
     with:
       branch: ${{ github.head_ref }}
       event_number: ${{ github.event.number }}
--- a/.github/workflows/deploy-pr-preview.yml
+++ b/.github/workflows/deploy-pr-preview.yml
@@ -9,16 +9,24 @@ on:
     paths:
       - "docs/sources/**"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ inputs.repo }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   deploy-pr-preview:
+    if: "!github.event.pull_request.head.repo.fork"
     uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@main
-    if: ${{ ! github.event.pull_request.head.repo.fork }}
     with:
-      sha: ${{ github.event.pull_request.head.sha }}
       branch: ${{ github.head_ref }}
       event_number: ${{ github.event.number }}
+      sha: ${{ github.event.pull_request.head.sha }}
+      sources: |
+        [{
+            "index_file": null,
+            "relative_prefix": "/docs/writers-toolkit/",
+            "repo": writers-toolkit",
+            "source_directory": "docs/sources",
+            "website_directory": "content/docs/writers-toolkit"
+        }]
       title: ${{ github.event.pull_request.title }}
-      repo: writers-toolkit
-      website_directory: content/docs/writers-toolkit
-      relative_prefix: /docs/writers-toolkit/
-      index_file: false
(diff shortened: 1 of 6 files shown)


docs/sources/_index.md
```diff
@@ -39,3 +39,5 @@ If it's your first time using the guide, start with the [Get started](https://gr
 
 Writers' Toolkit is open source and available at [`grafana/writers-toolkit`](https://github.com/grafana/writers-toolkit).
 If you have questions, or feedback on how to improve this documentation, [open an issue](https://github.com/grafana/writers-toolkit/issues/new) and help make this an even better resource.
+
+Test deploy preview
```
robbymilo, line 43: I have not dug into the code yet but at first glance I am not seeing this change on the deploy preview.

jdbaldry: 🤦 
Of course! I don't actually check out the PR branch. I need to fix that in the script...

jdbaldry: It's there now: https://deploy-preview-writers-toolkit-1014-zb444pucvq-vp.a.run.app/docs/writers-toolkit/#:~:text=even%20better%20resource.-,Test%20deploy%20preview,-Was%20this%20page

.github/workflows/deploy-preview.yml
```diff
@@ -12,24 +60,12 @@ on:
       event_number:
         required: true
         type: string
-      title:
```
robbymilo, line 15: Am I correct in assuming we will need to update each workflow file, and that there is not a way to support both? Could the original inputs be accepted, and the `sources` overrides them? Bad idea?

deploy-preview/build
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env bash
+
+set -euf -o pipefail
+
+if [[ -n "${RUNNER_DEBUG+x}" ]]; then
+  set -x
```
robbymilo, line 6: Neat!

deploy-preview/urls
robbymilo, line 1: I like that we are outputting all the URLs here, nice bonus.

.github/workflows/deploy-preview.yml
```diff
@@ -204,16 +227,26 @@ jobs:
         with:
           name: deploy_preview
           status: "${{ job.status }}"
-          url: "${{ steps.deploy.outputs.url }}${{ inputs.relative_prefix }}"
+          url: "${{ steps.deploy.outputs.url }}${{ (fromJSON(inputs.sources))[0].relative_prefix }}"
```
robbymilo, line 230: I think only a single URL here is fine but it might cause some confusion.

robbymilo (commented):
Looks good, just a question on updating each repo's workflow file.

.github/workflows/deploy-preview.yml
jdbaldry: Yeah I meant to check if I can put multiple URLs into the status check but I don't think I can

.github/workflows/deploy-preview.yml
jdbaldry: That might be the best way to roll this out actually.

The other thought was using one of our other branches (like this one or `robbymilo/deploy-preview`) so that we can kind of blue/green deploy changes.

In this case, I can probably modify the script to support both so we can smoothly migrate from the old to the new.

Will give you a ping after I've done that.

robbymilo (approved):
Looks good, great improvement!

.github/workflows/deploy-pr-preview.yml
```diff
@@ -11,14 +11,19 @@ on:
 
 jobs:
   deploy-pr-preview:
-    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@main
-    if: ${{ ! github.event.pull_request.head.repo.fork }}
+    if: "!github.event.pull_request.head.repo.fork"
+    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@jdb/deploy-preview
```
jdbaldry, line 15:
```suggestion
    uses: grafana/writers-toolkit/.github/workflows/deploy-preview.yml@main
```

PR #1058, opened 2025-02-19
Add Kotlin and Gunicorn product names to Vale dictionary

grafsean:


jdbaldry (commented):
Adding some typo swaps for better suggestions

jdbaldry (approved):

PR #1098, opened 2025-03-17
Document mount links behavior

jdbaldry:
I've decided to refer to the feature as "automatic mount links".

This documentation focuses on the guidance to a writer and tries to omit any unnecessary implementation detail.
The added note admonition covers the implementation of the behavior and leaves it up to the Docs Platform team to manage.


docs/sources/write/links/_index.md
```diff
@@ -35,7 +35,27 @@ Although these other types of links still function, replace them with one of the
 
 ## Link from source content that's reused as multiple pages
 
-Use `ref` URIs.
+Use a full URL and ensure the mount target has `mount_links: true` in its front matter.
+The `mount_links` front matter tells Hugo to link to mounted content in the current project if it exists.
+
+During link rendering, Hugo checks if the link destination to a mounted source.
```
JStickler, line 41:
```suggestion
During link rendering, Hugo checks if the link destination is a mounted source.
```

docs/sources/write/links/_index.md
```diff
@@ -35,7 +35,27 @@ Although these other types of links still function, replace them with one of the
...
+If it's mounted, then Hugo checks if the current page is in the same project as the mount target.
+If it's in the same project, then Hugo replaces the link destination with the mount target.
```
JStickler, line 43: I think "it is" would be clearer here than "it's".

docs/sources/write/links/_index.md
```diff
@@ -35,7 +35,27 @@ Although these other types of links still function, replace them with one of the
...
+If it isn't mounted, the mount target isn't in the same project, then Hugo uses the link as normal.
```
JStickler, line 44: This sentence is very confusing.

jdbaldry: It's missing a word! Does this help?

```suggestion
If it isn't mounted, and the mount target isn't in the same project, then Hugo uses the link as normal.
```

JStickler: Yes, now it makes sense.

docs/sources/write/links/_index.md
```diff
@@ -35,7 +35,27 @@ Although these other types of links still function, replace them with one of the
...
 ## Link from source content that's reused as multiple pages
 
-Use `ref` URIs.
+Use a full URL and ensure the mount target has `mount_links: true` in its front matter.
+The `mount_links` front matter tells Hugo to link to mounted content in the current project if it exists.
+
+During link rendering, Hugo checks if the link destination is a mounted source.
+If it's mounted, then Hugo checks if the current page is in the same project as the mount target.
+If it is in the same project, then Hugo replaces the link destination with the mount target.
+If it isn't mounted, and the mount target isn't in the same project, then Hugo uses the link as normal.
```
jdbaldry, line 44: I guess this should actually be an `or`?

```suggestion
If it isn't mounted, or the mount target isn't in the same project, then Hugo uses the link as normal.
```

clayton-cornell:
Is `mount_links: true` something we add to the cascaded frontmatter on the root `_index.md` within each doc repo?

jdbaldry:
> Is `mount_links: true` something we add to the cascaded frontmatter on the root `_index.md` within each doc repo?

Only the repositories that are on the "target" side of mounts. Things like Grafana Cloud, which mounts in docs from Grafana and many other places. Also GEM which mounts Mimir docs.

docs/sources/write/links/_index.md
```diff
@@ -35,7 +35,27 @@ Although these other types of links still function, replace them with one of the
 
 ## Link from source content that's reused as multiple pages
 
-Use `ref` URIs.
+Use a full URL and ensure the mount target has `mount_links: true` in its front matter.
```
imatwawana, line 38: I'm not clear on what file the target would be. Maybe using the alerting example below, indicate what the target would be here, and where specifically one would check this. If this isn't really something you want every writer checking then maybe this should be in more of a setup section?

docs/sources/write/links/_index.md
```diff
@@ -35,7 +35,27 @@ Although these other types of links still function, replace them with one of the
...
+During link rendering, Hugo checks if the link destination is a mounted source.
+If it's mounted, then Hugo checks if the current page is in the same project as the mount target.
+If it is in the same project, then Hugo replaces the link destination with the mount target.
+If it isn't mounted, and the mount target isn't in the same project, then Hugo uses the link as normal.
+
+### Mount links example
+
+The Grafana Labs website configuration mounts the open source Grafana Alerting documentation into the Grafana Cloud documentation.
+
+From the Grafana Alerting introduction page, the link `[alert rule](/docs/grafana/<GRAFANA_VERSION>/alerting/fundamentals/alert-rules/)` has different destinations in the open source and cloud versions of the page.
```
imatwawana, line 50: The directions above say to use a full URL, but that doesn't seem to be what's in this example.

docs/sources/write/links/_index.md
imatwawana, line 44: I'm not too concerned about this being here, but it definitely falls into information I don't need as a person using these.

docs/sources/write/links/_index.md
```diff
@@ -35,7 +35,27 @@ Although these other types of links still function, replace them with one of the
...
+In the open source version, the link stays the same.
+For Grafana Cloud, Hugo changes the link destination from [`/docs/grafana/next/alerting/fundamentals/alert-rules/`](/docs/grafana/next/alerting/fundamentals/alert-rules/) to [`/docs/grafana-cloud/alerting-and-irm/alerting/fundamentals/alert-rules/`](/docs/grafana-cloud/alerting-and-irm/alerting/fundamentals/alert-rules).
```
imatwawana, line 52:
```suggestion

- In the open source version, the link stays the same.
- For Grafana Cloud, Hugo changes the link destination from [`/docs/grafana/next/alerting/fundamentals/alert-rules/`](/docs/grafana/next/alerting/fundamentals/alert-rules/) to [`/docs/grafana-cloud/alerting-and-irm/alerting/fundamentals/alert-rules/`](/docs/grafana-cloud/alerting-and-irm/alerting/fundamentals/alert-rules).

```

imatwawana (commented):
This looks pretty good! I left a few questions and made a suggestion.

docs/sources/write/links/_index.md
```diff
@@ -35,7 +35,27 @@ Although these other types of links still function, replace them with one of the
 
 ## Link from source content that's reused as multiple pages
 
-Use `ref` URIs.
+Use a full URL and ensure the mount target has `mount_links: true` in its front matter.
+The `mount_links` front matter tells Hugo to link to mounted content in the current project if it exists.
```
tacole02, line 39: I didn't realize this was a necessary step. Is this something that needs to be set for every mounted page? Also, I think an example of what this looks like in the front matter would be really helpful.

docs/sources/write/links/_index.md
tacole02: I disagree. When I was working on this for Mimir/GEM it definitely helped me to think of it in terms of Hugo asking "Does this target URL also exist in this project? If yes, substitute for that topic."

docs/sources/write/links/_index.md
tacole02: Good call out!

tacole02 (approved):
Looks really good, @jdbaldry ! I had one outstanding question about what you set up in the front matter.

docs/sources/write/links/_index.md
jdbaldry: Good point, I'll just refer to URLs.

I think we've coined the term full URL to refer to a URL with scheme (like `https://`), hostname (like `grafana.com`), and URL pathinfo (like `/docs/grafana/`). In our internal links, we only ever need the pathinfo.

docs/sources/write/links/_index.md
jdbaldry: It's something we need for every mount target, that is, a place that has content mounted into it. At the moment, that includes Grafana Cloud, GEM, GEL, and GET.

I can provide an example of using cascading front matter to set this as a default for all pages in a project.

docs/sources/write/links/_index.md
```diff
@@ -35,7 +35,39 @@ Although these other types of links still function, replace them with one of the
 
 ## Link from source content that's reused as multiple pages
 
-Use `ref` URIs.
+Use a URL.
```
imatwawana, line 38:
```suggestion
Use a URL, including version interpolation if applicable.
For example:

`/docs/grafana/<GRAFANA_VERSION>/alerting/fundamentals/alert-rules/`
```

docs/sources/write/links/_index.md
imatwawana: I think that makes sense. I've made a suggestion above just to give people some sense of the format without having to read the example.

imatwawana (approved):
[One last suggestion](https://github.com/grafana/writers-toolkit/pull/1098#discussion_r2001886886), but otherwise this looks good to me!

docs/sources/write/links/_index.md
```diff
@@ -64,19 +96,19 @@ It can include hyphens (`-`).
 Hugo looks up _`<KEY>`_ in the value for the `refs` field in the page's front matter.
 If there is no _`<KEY>`_ in the `refs` field, or there is no `refs` field in the front matter, Hugo logs a build error.
 
-### Front matter
+#### Front matter
 
-{{< docs/shared source="writers-toolkit" lookup="refs-example.md" leveloffset="+2" >}}
+{{< docs/shared source="writers-toolkit" lookup="refs-example.md" leveloffset="+3" >}}
```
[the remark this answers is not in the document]
jdbaldry: Updated in 437ed24244f7963a248cf261ceebdf040a64fb85

imatwawana:
Looks great!

chri2547 (approved):

PR #1133, opened 2025-04-05
Remove references to `doc-validator`

jdbaldry:
The tool is no longer that useful as most of the good rules are in Vale.


robbymilo (commented):
Do we want to include aliases for the removed files?

jdbaldry:
> Do we want to include aliases for the removed files?

Mm, yeah, I only thought about that briefly. I think it may be worth redirecting to Vale stuff even though there isn't a 1-1 mapping

robbymilo (approved):

PR #1159, opened 2025-04-22
Document when to use `shared` or `docs/copy` shortcodes

jdbaldry:


robbymilo (approved):
Makes sense.

PR #1179, opened 2025-05-09
Add step that checks for binary assets

jdbaldry:


robbymilo (approved):
This is great, thank you!

PR #1197, opened 2025-06-18
Add Yugabyte to dictionary

aangelisc:
Adding Yugabyte to the dictionary to fix the failure in grafana/website#26316

### DIFF
--- a/vale/dictionary/y.jsonnet
+++ b/vale/dictionary/y.jsonnet
@@ -1,4 +1,5 @@
 local word = import './word.jsonnet';
 [
   word.new('YAML', '', 'noun') { abbreviation: true, elaboration: "YAML Ain't Markup Language", established_abbreviation: true },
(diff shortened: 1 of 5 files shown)


vale/dictionary/y.jsonnet
```diff
@@ -1,4 +1,5 @@
 local word = import './word.jsonnet';
 [
   word.new('YAML', '', 'noun') { abbreviation: true, elaboration: "YAML Ain't Markup Language", established_abbreviation: true },
+  word.new('Yugabyte', '', 'noun') { product: true },
```
clayton-cornell, line 4: Should this actually be `YugabyteDB`?  It'd be nice to include a definition of YugabyDB here as well

For reference, https://www.yugabyte.com/ uses YuagbyteDB for the product and Yugabyte as the company name.

PR #1206, opened 2025-07-28
Add documentation CI step that checks for renamed files containing the shared shortcode

jdbaldry:
Will drop some commits before merging that were used to test the behavior.

Relates to https://raintank-corp.slack.com/archives/C07R2REUULS/p1752245272948599?thread_ts=1752244152.551779&cid=C07R2REUULS


robbymilo (approved):
This works, great idea!

PR #1209, opened 2025-07-31
fix table on structure page

robbymilo:
It currently has broken HTML which causes the page to break (https://grafana.com/docs/writers-toolkit/structure/): 
Screenshot From 2025-07-31 14-59-41


jdbaldry (approved):
Thanks!

PR #1220, opened 2025-09-05
Clarify explanations of current and parent directory elements in Front Matter

idastambuk:
I found the explanation to be quite confusing. This is an attempt to make it clearer, though further suggestions are welcome!

thanks @njvrzm for the suggestions

### DIFF
--- a/docs/sources/write/front-matter/index.md
+++ b/docs/sources/write/front-matter/index.md
@@ -99,9 +99,9 @@ To determine the relative alias, you must first understand the meaning of the cu
 
 For an alias in the page `/docs/grafana/latest/alerting/manage-notifications/`:
 
-- The current directory element (`.`) refers to the directory that contains the current page's directory
+- The current directory element (`.`) refers to the directory that contains the current page.
 
-  In this example, this is the page `/docs/grafana/latest/alerting/`.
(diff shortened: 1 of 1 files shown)


docs/sources/write/front-matter/index.md
```diff
@@ -99,11 +99,11 @@ To determine the relative alias, you must first understand the meaning of the cu
 
 For an alias in the page `/docs/grafana/latest/alerting/manage-notifications/`:
 
-- The current directory element (`.`) refers to the directory containing the current page, _not_ the directory of the current page.
+- The current directory element (`.`) refers to the directory that contains the current page's directory
```
jdbaldry, line 102: What do you think about keeping the `_not_...`? I've noticed a lot of people expect the `.` to mean the current directory like in shells like Bash.


```suggestion
- The current directory element (`.`) refers to the directory that contains the current page's directory, _not_ the directory of the current page.
```

jdbaldry (commented):
Yeah, this reads much better thank you.

docs/sources/write/front-matter/index.md
idastambuk: Accepted the suggestion, thank you! 

Should the zizmor check on this be failing?

PR #1242, opened 2025-10-24
Add information about the more HTML comment

jtvdez:
Adds an entry in shortcodes for `<!--more-->` and information on when to use it in the section over contributing to what's new.


jtvdez:
@jdbaldry using `<!--more-->` in it's description breaks the preview I get from `make docs` locally somewhat (getting rid of the backticks around the first instance of it actually leaves the rest of the note fairly intact but then there's a weird line break and you can't see the initial instance at all. Any suggestions? 😄 also I doubt we'd consider this properly a shortcode, more of a tag really, but this seemed like the best place to tell people that we can use it in what's new posts? Happy to make adjustments as needed, could even be contained wholly within the contributing to what's new section, but I didn't wanna go off the rails with explanation there if it can be avoided

PR #1365, opened 2026-03-27
Add multi-tenancy word with swap options

knylander-grafana:
Adding this to make things consistent. We use this term a lot in the Tempo docs. 

The Google developer documentation style guide -- which the Grafana Writers' Toolkit defers to for terms not in its own word list -- explicitly says:

multi-tenancy: Hyphenate. We generally prefer to close prefixed words, but this is an exception because it's an established term.

- [X] I've used a relevant pull request (PR) title.


jdbaldry (approved):
Thank you! Sorry it took a while for this to get reviewed

PR #1377, opened 2026-04-16
docs: update frontmatter topic

JStickler:
Adding an example alias for Grafana Cloud, since the format is different and I'm always getting confused.

Also adding documentation for `_build` because it doesn't appear to be documented anywhere except for Grot Guides so I never can find it when I'm preparing private preview docs.

- [x] I've added an entry to the [What's new](https://github.com/grafana/writers-toolkit/blob/main/docs/sources/whats-new.md) page (only required for notable changes).

### DIFF
--- a/docs/sources/whats-new.md
+++ b/docs/sources/whats-new.md
@@ -18,6 +18,14 @@ weight: 175
 
 This page provides a summary of notable changes to Writers' Toolkit guidance.
 
+## April 2026
+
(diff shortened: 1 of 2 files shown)


knylander-grafana (approved):
Thank you for adding these updates. Not sure if you ened to add  a skip statement so the _build and Unversioned are ignored.

PR #1388, opened 2026-05-12
Tighten up docs validation policy

jdbaldry:
The policy bot rule for documentation validation was too broad.
It was gated only on `.github/workflows/validate-documentation.yml` existing, which is true for nearly every PR, so policy bot always required the documentation workflow result even when no documentation files changed.

This change restricts the rule to PRs that modify `docs/sources/**` or `vale/**`.
It also closes a bypass where a docs PR could delete `.github/workflows/validate-documentation.yml` to avoid the requirement, by separately requiring that workflow file to remain present when documentation changes.


knylander-grafana (approved):
Trusting you on this, Jack :)

PR #1394, opened 2026-05-14
Add async word to the dictionary

evictorero:
Adding `async` after the bot suggested it [here](https://github.com/grafana/grafana/pull/124938#discussion_r3243629312) 


clayton-cornell:
I wonder here if it's better to say "use the full term  `asynchronous` instead of the short form"?

clayton-cornell:
I'm closing this PR.

We should default to using the full term. So in this case, asynchronous instead of async.

PR #1424, opened 2026-06-17
Add 'textbox' entry to the dictionary

imatwawana:
Added a new entry for 'textbox' with its description.

### DIFF
--- a/vale/dictionary/t.jsonnet
+++ b/vale/dictionary/t.jsonnet
@@ -6,6 +6,7 @@ local word = import './word.jsonnet';
   word.new('templated', '', 'adjective'),
   word.new('Team Sync', '', 'noun') { description: 'Sync your auth providers teams and teams in Grafana.', product: true },
   word.new('Tempo', '', 'noun') { product: true, swaps: { tempo: 'Tempo' } },
+  word.new('textbox', '', 'noun') { description: 'A type of variable that provides a free text input field with an optional default value.' },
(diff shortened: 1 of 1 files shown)


vale/dictionary/t.jsonnet
```diff
@@ -6,6 +6,7 @@ local word = import './word.jsonnet';
   word.new('templated', '', 'adjective'),
   word.new('Team Sync', '', 'noun') { description: 'Sync your auth providers teams and teams in Grafana.', product: true },
   word.new('Tempo', '', 'noun') { product: true, swaps: { tempo: 'Tempo' } },
+  word.new('textbox', '', 'noun') { description: 'A type of variable that provides a free text input field with an optional default value.' },
```
clayton-cornell, line 9: Is this for the actual variable (where we'd wrap it in inline code formatting) ? or what we will call the UI element?

Wikipedia says "text box" (with a space) for reg  text
https://en.wikipedia.org/wiki/Text_box

Microsoft also says "text box" https://learn.microsoft.com/en-us/previous-versions/windows/desktop/bb246424(v=vs.85)

imatwawana: It's UI element that sometimes gets lower cased when talked about in general. So you might get the following sentences:

_Select the **Textbox** variable type.
When you add a textbox variable, it displays a free form field etc._

jdbaldry (approved):

jdbaldry:
Thank you!

PR #1426, opened 2026-06-22
Fix CVEs relating to undici and jsdiff

jdbaldry:
Most are resolved by the removal of the disabled add-to-docs workflow code but the vale action also benefits from the core bump to update undici


JStickler (approved):
LGTM

robbymilo (approved):

PR #1431, opened 2026-06-23
fix determination of latest vs next dir for deploy previews

robbymilo:
for https://github.com/grafana/website/issues/29598

Currently all PRs are using latest, this should fix that. See https://github.com/grafana/grafana/pull/127069


jdbaldry (dismissed):
Seems like a pretty nice simplification in the logic to be honest

jdbaldry (approved):

PR #1436, opened 2026-06-29
update link checker image

robbymilo:
Whoops, looks like we were using two different versions of the image.


jdbaldry:
Dang, I broke the policy-bot policy

jdbaldry (approved):

PR #1444, opened 2026-07-03
Fix handling of empty ref fields

jdbaldry:
Broken by Commit 045843df "Add support for external repo sources." (#1393)

Fix forward because a revert would also require a review from another member the @grafana/docs-tooling team which is currently not possible until Tuesday.


irenerl24 (approved):

nate-walters-grafana (commented):
Looks good to me

PR #1447, opened 2026-07-09
Fix some reported issues with the Makefile

jdbaldry:
- Fix typo in `--no-builtin-rule` to `--no-builtin-rules`.
  GNU Make accepts unambiguous prefixes, but the fix is still correct.
- Removed the dead code `docs-rm` target.
- Reworded the `make-docs` target description to better reflect the behavior.


nate-walters-grafana (approved):
Looks good!

PR #1449, opened 2026-07-13
Fix token matching for Hugo shortcodes missing post-argument whitespace

jdbaldry:
Extends TokenIgnores to also ignore Hugo shortcodes like `{{< figure src="..." alt="...">}}` that omit the space before the closing delimiter.


clayton-cornell (approved):
Do we "need" a trailing newline? Git is complaining :-)

jdbaldry:
Not a requirement but I don't like how prominent it is 😂

robbymilo (approved):

PR #1461, opened 2026-07-14
Also allow labeled trigger on deploy preview

jdbaldry:
Consumers of the reusable workflow should gate this appropriately


robbymilo (approved):
Love it

PR #1469, opened 2026-07-16
Limit comments to modified lines present in patch

jdbaldry:
Before, the GitHub comment system was used to filter the feedback. GitHub returns a 422 for comments on bits of the file that aren't present in the hunk and therefore can't be reviewed. However, the diff hunk includes some context for modifications and that means linting on surrounding lines can get through. 

This PR filters the prospective comments to only the modified lines present in the patch itself before posting so we shouldn't see comments like https://github.com/grafana/alloy/pull/6702/files#r3590942465 anymore

### DIFF
--- a/vale-action/report-vale.js
+++ b/vale-action/report-vale.js
@@ -33,10 +33,46 @@ var __importStar = (this && this.__importStar) || (function () {
     };
 })();
 Object.defineProperty(exports, "__esModule", { value: true });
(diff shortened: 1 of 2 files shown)


clayton-cornell (approved):
I think it's all OK. I haven't directly tested it to validate though

robbymilo (approved):

